PR #1, opened 2016-02-02
Fix link to turbolink-classic

gmile:
Not sure about it since `turbolink-classic` is a thing being referred to a lot in the PR, so close this at will.


sstephenson:
We’re still getting the house in order. :grin: [rails/turbolinks](https://github.com/rails/turbolinks) will be moving to [turbolinks/turbolinks-classic](https://github.com/turbolinks/turbolinks-classic) shortly.

gmile:
Ah! Got it.

PR #3, opened 2016-02-05
Trim whitespace

tompedals:
Noticed some trailing whitespace in one file while browsing around the code. Not important but thought I'd do a quick PR just to tidy it up.


sstephenson:
Thanks!

PR #5, opened 2016-02-05
Add `turbolinks:before-render` event

sstephenson:
The `turbolinks:before-render` event fires just after Turbolinks creates the new body element but before it imports permanent elements. The new body is exposed as `event.data.newBody`.

This gives the application and third-party libraries a chance to query or manipulate the current and new body elements just before the new body is installed.


packagethief:
Really nice. And cheers for tests! :+1:

PR #8, opened 2016-02-10
Add option to skip cache version in links

### LINKED ISSUE #7: rendering twice via cache + reloaded
Hi,
is there a way to off the render from cache?
because the cached version is rendered while its trying to load the latest version.
its not good when we have a form in place, whereby user may started to type in some input in the text field and the server seems to take sometime to load, and causes the form to entirely rerendered.

yes. im using git version of turbolinks 5
thanks

jmbejar:
The motivation of this PR is to explore a way to fix #7.

Upgrading an application to use this new version of Turbolinks, we found a few situations where loading the cached version when visiting a new page is not working fine. In our case, we would like to skip loading the cached version in two cases:

- We don't want to show data entered by users when visiting again a page with a form
- We also have permalinks to sections where a modal dialog is opened. The animation triggers twice in those cases.

I couldn't find in the API any way to deal with this cases. I'm aware that taking advantage of the `turbolinks:before-cache` event could be used to solve the form issue, e.g. clearing the form at this point. However, I didn't know how to fix the animation problem without adding this option.
Even for the form case, I feel like having a way to skip the cached version would be a much simpler solution.

In case of adding this option, the name of the data attribute can be further discussed.

What do you think about this addition?

### DIFF
--- a/src/turbolinks/controller.coffee
+++ b/src/turbolinks/controller.coffee
@@ -87,10 +87,14 @@ class Turbolinks.Controller
   getCachedSnapshotForLocation: (location) ->
     @cache.get(location)
 
+  shouldCacheSnapshot: ->
+    @getSetting("cache-control") isnt "no-cache"
+
   cacheSnapshot: ->
-    @notifyApplicationBeforeCachingSnapshot()
-    snapshot = @view.getSnapshot()
-    @cache.put(@lastRenderedLocation, snapshot)
+    if @shouldCacheSnapshot()
+      @notifyApplicationBeforeCachingSnapshot()
+      snapshot = @view.getSnapshot()
+      @cache.put(@lastRenderedLocation, snapshot)
 
   # Scrolling
 
--- a/src/turbolinks/snapshot.coffee
+++ b/src/turbolinks/snapshot.coffee
@@ -35,6 +35,9 @@ class Turbolinks.Snapshot
   getTemporaryHeadElements: ->
     @getTemporaryHeadElementSet().getElements()
 
+  isPreviewable: ->
+    @getCacheControlValue() isnt "no-preview"
+
   # Private
 
   getTrackedHeadElementSet: ->
@@ -55,3 +58,6 @@ class Turbolinks.Snapshot
   getHeadElementSet: ->
     @headElementSet ?= new Turbolinks.ElementSet @head.childNodes
 
+  getCacheControlValue: ->
+    [..., element] = @head.querySelectorAll("meta[name='turbolinks-cache-control']")
+    element?.getAttribute("content")
--- a/src/turbolinks/visit.coffee
+++ b/src/turbolinks/visit.coffee
@@ -46,9 +46,10 @@ class Turbolinks.Visit
       @request.send()
 
   getCachedSnapshot: ->
-    snapshot = @controller.getCachedSnapshotForLocation(@location)
-    return if @location.anchor? and not snapshot?.hasAnchor(@location.anchor)
-    snapshot
+    if snapshot = @controller.getCachedSnapshotForLocation(@location)
+      if not @location.anchor? or snapshot.hasAnchor(@location.anchor)
+        if @action is "restore" or snapshot.isPreviewable()
+          snapshot
 
   hasCachedSnapshot: ->
     @getCachedSnapshot()?


packagethief:
Hi Jorge --

Thanks for taking the time to explore this. I agree that there should be some way to control page caching behavior. Before we discuss that, I wanted to address the two cases you mentioned, as I think you can solve them without needing explicit caching control.

> We don't want to show data entered by users when visiting again a page with a form

As you've noted, the `turbolinks:before-cache` event is the appropriate time to reset forms and make any other preparations before the DOM is stored. You'll likely want to do this even with an option to disable previews, since the same cached snapshot will be shown on a `restore` when browsing by history.

> We also have permalinks to sections where a modal dialog is opened. The animation triggers twice in those cases.

You should be able to wait for `turbolinks:load` before applying behavior. This is fired once after each visit, whether or not a preview was shown. If that's not possible (say, because you're using `MutationObserver` and are notified at each appearance of your element in the DOM), you can detect whether Turbolinks is displaying a preview by checking for the `data-turbolinks-preview` attribute on the `<html>` element. If you have behavior that you don't want applied to a preview, you can avoid installing it. (I wrote more about the preview DOM in [the comments on Issue #7](https://github.com/turbolinks/turbolinks/issues/7#issuecomment-182568753)).


All that said, an option to control caching behavior would be a welcome addition – I know it's come up in our use at Basecamp. I'm not sure that a link annotation or an option to `Turbolinks.visit()` is the way to go, though. When would you want to disable the preview only sometimes for a location?

I think this option should live at the page level. In other words, pages should be able to declare their caching behavior. I think a `<meta>` element is appropriate for such a declaration.

Perhaps something like `<meta name="turbolinks-show-cache" content="...">`, where `content` can be one of `never`, `only-restore`, `only-preview`, or `always`. The implied default would be `always`.

To declare that a cached page should only be shown for a restore, but not for a preview (where its eventual replacement might cause a visual glitch), you'd do the following:

```html
<head>
  <meta name="turbolinks-show-cache" content="only-restore">
</head>
```

I'm not sure about the names of everything, but that's the general idea. What do you think?

jmbejar:
Hey @packagethief, thanks for the feedback!

I agree that most of the issues mentioned in this thread and #7 can be solved by using `turbolinks:before-cache` and `turbolinks:load` events, or checking the presence of `data-turbolinks-preview`.

I only have one case that I can not solve with these options: the issue that I mentioned in #7 about the sidebar values that sometimes shows an inaccurate value navigating to a different section, ending up in a visual glitch. Since `data-turbolinks-permanent` is not an option in our specific situation, probably the best way to fix the problem with the current API is clearing the cache `Turbolinks.clearCache()` as you suggested (with the associated cost of losing the cache when navigating back).

Although these solutions are feasible, I happy to hear that you find valuable the possibility of opt out the preview cache. I see this addition as a huge simplification, at least in our specific app (I hope it would be the same for many apps!).

> All that said, an option to control caching behavior would be a welcome addition – I know it's come up in our use at Basecamp. I'm not sure that a link annotation or an option to Turbolinks.visit() is the way to go, though. When would you want to disable the preview only sometimes for a location?
> 
> I think this option should live at the page level. In other words, pages should be able to declare their caching behavior. I think a <meta> element is appropriate for such a declaration.

I don't have a strong opinion here. In fact, having the option in the link as I implemented in this PR is not the perfect fit for our app either. That's because I ended up adding the `data-turbolinks-skip-preview` to almost every link :tired_face:.

When I realized about this issue (after opening this PR), I wished to have a global setting, e.g. `Turbolinks.turbolinksCacheMode = ...` instead of marking individual links.

> Perhaps something like <meta name="turbolinks-show-cache" content="...">, where content can be one of never, only-restore, only-preview, or always. The implied default would be always.

I like this idea! It's similar to defining it as a global setting but scoped to pages.
I'm not sure if there is any real use case where you want to use the cache "only for preview", but I'm OK with the overall idea :+1:

Let me rework this PR using the `meta` attribute and test against our app. I will share the results.

jmbejar:
Hi @packagethief, please take a look now. I've tested this new implementation relying on a meta attribute as we discussed last week.

This solution solves the issues that we have in our application without adding extra complexity or hacks.

Regarding the implementation, I'm not sure if we should add tests for this changes. I noticed there are some tests but I didn't find a easy way to add tests for this addition. What would you do in this case? Any recommendation?

Thanks

packagethief:
Hey @jmbejar, thanks a lot for working on this.

During a visit, I think we should ask the _cached_ page whether it allows itself to be shown for the requested action–not the current page. That way, pages are responsible for declaring their own caching behavior.

To do this we'll need to load the snapshot from cache and query its `<head>` for the setting.

Perhaps `Visit` can perform this check as part of `getCachedSnapshot()`. It already checks whether the snapshot should be used for anchored locations. That could be extracted, along with a check for the cache setting, to something like `shouldUseCachedSnapshot(snapshot)`.

Then `hasCachedSnapshot` will be accurate, and `getCachedSnapshot` can be reduced to:

```coffee
getCachedSnapshot: ->
  snapshot = @controller.getCachedSnapshotForLocation(@location)
  snapshot if snapshot and @shouldUseCachedSnapshot(snapshot)
```

I also think we should include support for a "never" option. If you have a page that isn't well-suited to being cached, you won't want it shown for previews or restores. In that case it would probably be simplest  to avoid caching the page altogether.

What do you think?

jmbejar:
> During a visit, I think we should ask the cached page whether it allows itself to be shown for the requested action–not the current page. That way, pages are responsible for declaring their own caching behavior.

Yeah, that's a valid point. In our specific app it doesn't make a big difference because we decided to mark all pages with `only-restore` for now, however I think you're totally right. I'll do the change.

> I also think we should include support for a "never" option. If you have a page that isn't well-suited to being cached, you won't want it shown for previews or restores. In that case it would probably be simplest to avoid caching the page altogether.

Right, so all the possibilities are covered :+1:

packagethief:
Hey Jorge,

Before you try implementing this further, I was just discussing it with @sstephenson and we're thinking that the original suggestion of options like "only-preview" and "only-restore" isn't quite right. For one, I can't come up with a good reason why you'd want "only-preview".

It seems to me that most common cases are: 1) don't cache this page ever, and 2) don't use a cached copy of this page for a preview. These are the only two cases I think we ought to support.

I know I suggested it originally, but the name "turbolinks-show-cache" doesn't seem quite right either. It doesn't do a very good job of communicating that this is a per-page option for declaring caching directives.

To that end, here's a revised proposal:

I think it would be cute if we paid homage to the HTTP Cache-Control header (especially since it can also be specified in HTML with a `<meta>` tag). We'd name our tag "turbolinks-cache-control", and accept two directives: "no-cache" and "no-preview".

To prevent the page from ever being cached by Turbolinks:

```html
<meta name="turbolinks-cache-control" content="no-cache">
```

Or, to prevent a cached copy of the page from being used for a preview during an [application visit](https://github.com/turbolinks/turbolinks#application-visits):

```html
<meta name="turbolinks-cache-control" content="no-preview">
```

In addition to being clearer, I think this will simplify the implementation. Snapshots can be responsible for knowing whether they're cacheable (`isCacheable`) or previewable (`isPreviewable`). Then it's just a matter of asking a snapshot these questions before we use it. Pages declared "no-cache" wouldn't be saved to the cache at all.

What do you think?

jmbejar:
@packagethief I can say that this new proposal will cover the essential cases. I agree that there should not be cases where the "only-preview" option would be useful.

I will follow this new ideas and get back to you if some doubt arises, however I think it should work fine :+1:

jmbejar:
Here is a new implementation following yesterday's feedback.

We have now two commits, the first one (https://github.com/jmbejar/turbolinks/commit/f1f694ac4d59accdd053b2baebd60239094c002f) adds the ability to prevent previews.
You can see the `isPreviewable` function added to `Snapshot` as we planned.

In the second commit (https://github.com/jmbejar/turbolinks/commit/072ce15474e9c564e30bf7c4a78b7104edb2a53e), I had problems implementing the `isCacheable` within `Snapshot`, it seems to fit better in the controller. This is because we should check if the page is cacheable right before notifying the app (triggering "before-cache" event), however I need to trigger those events before generating the snapshot:

```coffeescript
  cacheSnapshot: ->
    if @isCacheable()
      @notifyApplicationBeforeCachingSnapshot()
      snapshot = @view.getSnapshot()
      @cache.put(@lastRenderedLocation, snapshot)
```

Therefore, I should not wait for the snapshot generation to check the current page meta tag for "cache-control".

I think this implementation has the best of both worlds, with minimal code duplication (e.g. "cache-control" retrieval is implemented twice, in Snapshot and Controller)

src/turbolinks/visit.coffee
```diff
@@ -56,11 +56,12 @@ class Turbolinks.Visit
   loadCachedSnapshot: ->
     if snapshot = @getCachedSnapshot()
       isPreview = @shouldIssueRequest()
-      @render ->
-        @cacheSnapshot()
-        @controller.render({snapshot, isPreview}, @performScroll)
-        @adapter.visitRendered?(this)
-        @complete() unless isPreview
+      if snapshot.isPreviewable() or not isPreview
```
jmbejar: I feel this new condition made the whole `loadCachedSnapshot ` function a quite harder to follow. Let me think if we can refactor this.

src/turbolinks/controller.coffee
```diff
@@ -217,6 +218,9 @@ class Turbolinks.Controller
     root = @getSetting("root") ? "/"
     new Turbolinks.Location root
 
+  isCacheable: ->
```
packagethief: I think `shouldCacheSnapshot` might be a clearer name. I'd also move this method to the "Snapshot cache" section above, just after `cacheSnapshot`.

src/turbolinks/controller.coffee
```diff
@@ -217,6 +218,9 @@ class Turbolinks.Controller
...
+    @getSetting("cache-control") != 'no-cache'
```
packagethief: Minor style nit: we've been preferring `isnt` over `!=`, and double over single quotes elsewhere in the source.

```coffeescript
shouldCacheSnapshot: ->
  @getSetting("cache-control") isnt "no-cache"
```

src/turbolinks/snapshot.coffee
```diff
@@ -35,6 +35,9 @@ class Turbolinks.Snapshot
   getTemporaryHeadElements: ->
     @getTemporaryHeadElementSet().getElements()
 
+  isPreviewable: ->
+    @getCacheControlValue() != 'no-preview'
```
packagethief: Same style nit as above: prefer `isnt` and double quotes.

src/turbolinks/visit.coffee
packagethief: I think this check should be moved to `getCachedSnapshot`. `getCachedSnapshot` should return nothing if a usable snapshot can't be found. Just like it does for snapshots that lack a requested anchor.

```coffee
getCachedSnapshot: ->
  if snapshot = @controller.getCachedSnapshotForLocation(@location)
    if not @location.anchor? or snapshot.hasAnchor(@location.anchor)
      if @action is "restore" or snapshot.isPreviewable()
        snapshot
```

This way we won't need to change `loadCachedSnapshot` or `hasCachedSnapshot`.

src/turbolinks/controller.coffee
jmbejar: :+1: Apart from style preferences, I'm just releasing that `isnt` translates to `!==`, and that's much better than `!=`.

Will change

src/turbolinks/controller.coffee
jmbejar: :+1:

jmbejar:
@packagethief I've addressed all your feedback. Thanks again for all your help!

src/turbolinks/controller.coffee
```diff
@@ -87,10 +87,14 @@ class Turbolinks.Controller
   getCachedSnapshotForLocation: (location) ->
     @cache.get(location)
 
+  shouldCacheCacheable: ->
```
packagethief: I think you mean `shouldCacheSnapshot` :grin:

jmbejar: :dizzy_face: 

Fixing it now, and I'm also organizing better the PR commits (my last push introduced unnecessary back-and-forth in the code changes)

sstephenson:
Excellent work, @jmbejar. Thanks for seeing this through.

jmbejar:
@sstephenson @packagethief :+1:

PR #18, opened 2016-02-23
Open up HTTP request headers on Turbolinks.visit()

joemasilotti:
This PR gives the user an option to specify HTTP request headers when programmatically making a `Turbolink.visit()` request.

I am doing something similar with turbolinks-classic. I'm posting here for two reasons. Ideally to get this merged in so when we upgrade we can hit the ground running. And second, to ensure this doesn't open up any security (or other) holes.


sstephenson:
Thanks for the pull request. I’m curious what kind of request headers you need to set, and why you’d only want to set them on calls to `Turbolinks.visit` but not on link clicks.

joemasilotti:
I'm building a hybrid iOS app. Each time the user taps a link the request is cancelled and then "redone" via Turbolinks. In between these two phases some HTTP headers are added to set a variety of information. This ranges from the user's ID, their auth token, and other state-based meta data that is stored on the client.

joemasilotti:
Is there a better way to do what I am trying to accomplish, @sstephenson?

sstephenson:
Hey @joemasilotti,

I don’t see a good reason to expose this on `Turbolinks.visit`. I think you’d be better served by listening for the [`turbolinks:request-start` event](https://github.com/turbolinks/turbolinks/blob/93f56a107549ed00c5609fbc09190b6d77008d60/src/turbolinks/http_request.coffee#L54-L55) and setting any custom headers on the XHR there:

```js
document.addEventListener("turbolinks:request-start", function(event) {
  var xhr = event.data.xhr
  xhr.setRequestHeader("X-User-Id", "123")
  // ...
})
```

That way the headers apply to visits from link clicks too.

Closing this for now, but feel free to re-open if I’m missing something.

sstephenson:
BTW, I just realized the `request-start` documentation in the Events section doesn’t cover this use case. I’d be happy to accept a pull request that covers setting custom headers in the readme.

joemasilotti:
Ah, I wasn't aware that the XHR lives on the event. I just submitted [a small patch](https://github.com/turbolinks/turbolinks/pull/31) noting how to do this. I felt that it didn't fit well in the Events section, though. I'm happy to move it if you disagree. Thanks for the help!

PR #23, opened 2016-02-26
Use Element.set/removeAttribute() for IE10 compat. Fixes #21

### LINKED ISSUE #21: IE10 doesn't seem to work in 5.0.0.beta2
IE10 is showing a progress bar, Ajax request works fine, but content doesn't get replaced.

packagethief:
Switches to `Element.setAttribute()` and `removeAttribute()` in favor of using `Element.dataset`, which [isn't supported in IE < 11](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_data_attributes#Issues).

Fixes #21

### DIFF
--- a/src/turbolinks/view.coffee
+++ b/src/turbolinks/view.coffee
@@ -19,9 +19,9 @@ class Turbolinks.View
 
   markAsPreview: (isPreview) ->
     if isPreview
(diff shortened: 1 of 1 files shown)


sstephenson:
👍

PR #24, opened 2016-02-26
Flesh out using Custom Elements with Turbolinks

qrush:
Added a little example based on the conversation at jwo/turbolinks-exploration-google-charts#1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ You may be used to installing JavaScript behavior in response to the `window.onl
 In many cases, you can simply adjust your code to listen for the `turbolinks:load` event, which fires once on the initial page load and again after every Turbolinks visit.
 
 ```js
-document.addEventListener("turbolinks:load",  {
+document.addEventListener("turbolinks:load", function() {
   // ...
 })
 ```
@@ -156,7 +156,7 @@ Turbolinks saves a copy of the current page to its cache just before rendering a
 Listen for the `turbolinks:before-cache` event if you need to prepare the document before Turbolinks caches it. You can use this event to reset forms, collapse expanded UI elements, or tear down any third-party widgets so the page is ready to be displayed again.
 
 ```js
-document.addEventListener("turbolinks:before-cache",  {
+document.addEventListener("turbolinks:before-cache", function() {
   // ...
 })
 ```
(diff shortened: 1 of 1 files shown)


nateberkopec:
Cool approach and a solid example.

README.md
```diff
@@ -191,7 +191,32 @@ Turbolinks may not be the only source of page updates in your application. New H
 
 You can handle all of these updates, including updates from Turbolinks page loads, in a single place with the precise lifecycle callbacks provided by [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) and [Custom Elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Custom_Elements).
 
-In particular, these APIs give you callbacks when elements are attached to and removed from the document. You can use these callbacks to perform transformations and register or tear down behavior as soon as matching elements appear on the page, regardless of how they were added.
+In particular, these APIs give you callbacks when elements are attached to and removed from the document. You can use these callbacks to perform transformations and register or tear down behavior as soon as matching elements appear on the page, regardless of how they were added. For example, if you're checking for elements on the page like so:
+
+``` javascript
+document.addEventListener("turbolinks:load", function() {
+  if (document.getElementById('sparkline')) {
```
nateberkopec: Rather than hooking on `turbolinks:load`, couldn't you just attach an event delegator to the `html` element?

README.md
```diff
@@ -191,7 +191,32 @@ Turbolinks may not be the only source of page updates in your application. New H
...
+    drawSparkline();
+  }
+});
+```
+
+Instead, you could use Custom Elements to listen for this element. This is a great way to implement and hook in third party libraries with Turbolinks as well. You could a new element in your page like so:
```
kaspth: `add a` ;)

README.md
qrush: This is just an example of what is a "bad" way of working with TL. I'm not sure if any further elaboration is required.

README.md
qrush: Thanks, fixed in 54755f5.

packagethief:
Thanks for this, Nick.

As we discussed earlier, I don't think this example is a good fit for the main README. Custom elements are great, but they're also rather advanced and still require a polyfill to work. I don't necessarily think everyone should use them for everything, and I don't want to give them undue weight.

Let's move this and other "advanced" examples to the wiki.

Issue #26, opened 2016-02-28
Does Turbolinks have to check location's extension before visiting it?


akxcv:
Hello! I've got the following issue with Turbolinks: if I try to visit a URL like `/chris.martin`, Turbolinks doesn't perform a Visit and the whole page gets reloaded. If I remove the dot from this URL (`/chrismartin`), Turbolinks performs a Visit as it should.

In the Turbolinks source code:
```coffeescript
# controller.coffee
locationIsVisitable: (location) ->
  location.isPrefixedBy(@getRootLocation()) and location.isHTML()

# location.coffee
getExtension: ->
  @getLastPathComponent().match(/\.[^.]*$/)?[0]

isHTML: ->
  extension = @getExtension()
  extension is ".html" or not extension?
```

So my question is: Why does Turbolinks check URL's extension before visiting it? Maybe we need to add an optional attribute like `data-turbolinks-extension="false"`? Or is there a better way do solve this problem?

**EDIT**: Now that I think about it, this extension check makes sense (what if I request a raw .js file, for example?). But I think we need an option to disable it on specific links.

akxcv:
If `data-turbolinks-extension="false"` is a good solution to this problem, I'm ready to create a pull request.

akxcv:
#29

akxcv:
#29

packagethief:
Hey @akxcv, we discussed this last week in #15.

tldr; we think applications should be able to define what constitutes an "HTML URL." Until there's official API for that, you can [override Turbolinks.Location.prototype.isHTML](https://github.com/turbolinks/turbolinks/issues/15#issuecomment-186639159).

Issue #15, opened 2016-02-19
URLs with decimals do not get rendered


aaronvb:
This might be a unique case, but I found this difference when migrating an existing project from classic that uses decimals in the URL path. Turbolinks 5 seems to have dropped support for that.

Example use case:
````
/products/canon-85mm-f-1.2-portrait-lens
````

The above URL will not render and instead reload.

I found the difference in the regex used to parse the URL extension.

https://github.com/aaronvb/turbolinks/blob/master/src/turbolinks/location.coffee#L34

`/\.[^.]*$/`

Which will match`.2-portrait-lens` as the extension.

https://github.com/turbolinks/turbolinks-classic/blob/master/lib/assets/javascripts/turbolinks.coffee#L462

`/\.[a-z]+$/`

Which will not have a match.

Was this change intended?

I've created a branch using the classic regex and supporting tests: https://github.com/aaronvb/turbolinks/tree/fix_get_extension_for_decimals

sstephenson:
I think it’s more likely that Turbolinks Classic’s regex was a little too loose. 

I’m not aware of any rule, formal or informal, saying file extensions can’t start with digits. Two common counterexamples off the top of my head: `.7z` and `.3gp`.

In your case I’d suggest explicitly including a `.html` suffix in your links. I’m also open to discussing ways we can make this behavior customizable.

aaronvb:
Good points, I did not think about those types of file extensions.

I would still like to get this to work without appending `.html` in my URLs if possible. What do you think about an option to ignore the extension altogether? ie: `data-turbolinks-ignore-extension=true` Although I can't think of another reason to want to ignore the extension other than my case.

sstephenson:
Yeah, I don’t think a link annotation really makes sense here, but maybe we can come up with a way to make the [`isHTML` condition](https://github.com/turbolinks/turbolinks/blob/e870d0ea2dda7b68144f5bdcae46417d5754a01c/src/turbolinks/controller.coffee#L227-L228) configurable.

aaronvb:
If we don't use link annotation, the only way to configure isHTML would be to pass an option to the Turbolinks.visit() function, correct? I don't think that wouldn't apply in my case because I'm generating the URLs in the view.

sstephenson:
I’d prefer to make it globally configurable so your application can define its own rules for what constitutes an “HTML link.” 

[...]

PR #29, opened 2016-02-28
Added an option to disable location extension check.

akxcv:
Turbolinks will not check a location's extension before visiting it if `data-turbolinks-extension="false"` is specified.

Or, programmaticaly, location's extension check can be avoided like so:
```js
Turbolinks.visit("/jane.cooper", { shouldCheckExtension: false })
```


akxcv:
#26

packagethief:
Thanks @akxcv. I agree that URL extension eligibility should be customizable, but not on a per-visit basis. See #15 for more discussion.

akxcv:
Thank you for your explanation @packagethief. I should've read closed issues :grimacing:

PR #33, opened 2016-03-03
Expose clicked HTML anchor element to BrowserAdapter

joemasilotti:
The direct goal of this PR is to expose the clicked HTML anchor tag to the `BrowserAdapter`. Indirectly, this provides the Turbolinks Android project the [necessary scaffolding to receive the same information](https://github.com/turbolinks/turbolinks-android/pull/7).

The purpose of exposing an HTML element to the adapter is to keep more information/logic on the server. This is accomplished by embedding metadata in the node itself to provide context for a native client.

For example, setting `data-screen-title` on an anchor could be interpreted on the client to set the title of the next screen, before any page loading begins! Or setting `data-type-form` to have the client always use the same activity or view controller with form behavior, implemented with native Java or Swift/Objective-C.

More details can be found in [this issue](https://github.com/turbolinks/turbolinks-android/issues/6) I opened a few days ago which stemmed from a [conversation with @lateplate](https://twitter.com/joemasilotti/status/704366532613312512).


joemasilotti:
The most recent commit, [94d017a](https://github.com/joemasilotti/turbolinks/commit/94d017aca523e95b26280ef02f1be2ec0ae8d671), first finds the closest anchor tag to the clicked link. This is useful when you have a `<span>` or `` inside the anchor tag but the data attributes exist on the anchor tag.

joemasilotti:
Congratulations on the Turbolinks iOS release today! I'm excited to start digging in to that.

Are there any thoughts on this PR? If all goes well I will propose the same Android changes to the new iOS library.

packagethief:
Thanks for exploring this, Joe.

I appreciate the desire to provide context for native clients, but I don't think per-link metadata is the best approach. Clicking a link is only one way to initiate a Turbolinks visit. How would you provide context when performing a programmatic visit?

I think this type of configuration belongs at the URL level, not the link level.

Consider a mapping of URL patterns to configuration instructions. When a visit is proposed, whether it’s from a click or programmatically, you can match its URL to configuration and decide how to proceed. For example, you might map URL paths matching "*/edit" to configuration that specifies modal presentation.

If you keep this mapping in a manifest file on the server and have clients download it, you can modify it in one place, and share it across platforms.

This is a more flexible approach that doesn't require you to spread client-specific presentation instructions throughout your views. You can also use the same configuration when initiating a Turbolinks visit from a native client, and avoid needing to specify it in multiple places.

joemasilotti:
> How would you provide context when performing a programmatic visit?

My thought is to eventually provide all of the `data-` attributes in a hash via this API. That way, you could pass the same attributes to `visit()` and achieve the same functionality. Perhaps this would could be better explained through an additional commit, as the change is trivial.

> I think this type of configuration belongs at the URL level, not the link level.
> ...
> If you keep this mapping in a manifest file on the server and have clients download it, you can modify it in one place, and share it across platforms.

I don't disagree that this is a good idea; I can definitely see the benefits of sharing a manifest between platforms, as well. However, URL routing is only half of the features this PR addresses.

The other half addresses passing metadata to the client without having to wait for the page to load. I mentioned the simple example of passing the screen title in original comment. Because the hook is fired before the new screen is presented, there is more flexibility with how and what to actually present.

Our current app, which uses a modified versions of Turbolinks 3, uses this to pass object metadata to the native clients. Model IDs, names, and other properties are needed when transitioning from a hybrid screen to a fully native screen. For example, our *edit* screen is fully native but is linked to from a *show* view. The *edit* screen needs to know the name and other attributes upon instantiation.

We experimented with putting all of this in the URL but it can quickly become cumbersome. Especially when we keep certain display logic on the server and don't want the URL to be case-sensitive or escape lots of characters.

joemasilotti:
> My thought is to eventually provide all of the data- attributes in a hash via this API. That way, you could pass the same attributes to visit() and achieve the same functionality. Perhaps this would could be better explained through an additional commit, as the change is trivial.

I've taken a stab at this [on my fork](https://github.com/joemasilotti/turbolinks-ios/commit/1bed3d9b563424160bd780798362143280e13efd).

sstephenson:
Joe, I’m delighted that you’ve jumped in and made changes in all three libraries to get this working!

I’m afraid this just isn’t the direction we want to take the project, though. Our recommended approach is to pass all your URL routing through a single function in your app, and map the URL to its metadata in that function. That way you’re guaranteed to have consistent metadata—whether you’re following a Turbolinks visit, initiating a visit from a native view, or acting on an external notification.

As @packagethief explained, we ran into the same problem at Basecamp when we built our native apps. We solved it with a manifest file that maps URL patterns to metadata, like which view controller class to use and which background color the view should have.

Issue #40, opened 2016-03-08
Rails 5 respond_to (format.xml) unexpected behaviour


petermumford:
When using a **respond_to** block and if you have a `format.xml` it will always use this no matter what else you have even if you're expecting a `html` response. Please look at this issue on rails rails/rails#24035 to get a full description of the issue.

packagethief:
Thanks @petermumford. Confirmed. This is likely because [we set the Accept header to include "application/xml"](https://github.com/turbolinks/turbolinks/blob/master/src/turbolinks/http_request.coffee#L66). This seems unnecessary and can be removed.

I'll open a PR to remove it later unless someone beats me to it.

packagethief:
Fixed by removing "application/xml" from the Accept header in #41

PR #41, opened 2016-03-08
Remove 'application/xml' from Request Header

petermumford:
This is to fix issue #40 so `format.xml` from within a **respond_to** block doesn't override all other response types.


packagethief:
Thanks @petermumford!

PR #48, opened 2016-03-23
Start HTTP request before changing browser history

nateberkopec:
Tests still pass, LMK if this breaks things. 

Chrome Timeline *after* this patch - network requests begin 20ms earlier.

screen shot 2016-03-23 at 3 23 59 pm


sstephenson:
Thank you!

PR #49, opened 2016-03-23
Remove unneccessary clones

nateberkopec:
Tested on a 4,000-element DOM.

[Before and after](http://imgur.com/a/1JvnJ) - about ~50ms saved.

I'm not sure if these are really necessary - tests still pass, correct me if I'm wrong here.

### DIFF
--- a/src/turbolinks/controller.coffee
+++ b/src/turbolinks/controller.coffee
@@ -93,7 +93,7 @@ class Turbolinks.Controller
   cacheSnapshot: ->
     if @shouldCacheSnapshot()
       @notifyApplicationBeforeCachingSnapshot()
-      snapshot = @view.getSnapshot()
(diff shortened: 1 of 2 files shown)


packagethief:
Nice catch @nateberkopec! Thank-you!

nateberkopec:
@packagethief It seems to me that it might be more sensible to change `getSnapshot` to *not* clone by default, as that seems the more common use, and future users of `getSnapshot` will avoid making the same mistake in the future by requesting cloning only if they need it.

sstephenson:
I agree. Let’s change the default to `false`. (Think that was just a slip-up on my part.)

nateberkopec:
Updated. I think I got the logic correct - please double check.

sstephenson:
💝

PR #52, opened 2016-03-23
Eliminate clones when caching has been disabled

nateberkopec:
This is probably more CoffeeScript than I've ever written in my life, so let me know if this can be organized better.

If the new Snapshot isn't going to be cached, it's more memory-and-time efficient to just insert that Snapshot directly into the page rather than cloning it.

### DIFF
--- a/src/turbolinks/controller.coffee
+++ b/src/turbolinks/controller.coffee
@@ -88,7 +88,7 @@ class Turbolinks.Controller
     @cache.get(location)
 
   shouldCacheSnapshot: ->
-    @view.getCacheControlValue() isnt "no-cache"
+    @view.shouldCacheSnapshot()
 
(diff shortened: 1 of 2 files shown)


sstephenson:
Thanks very much for the patch, but I don’t think we want to go down the road of micro-optimizations for specific cases like this. At least not without convincing benchmarks from real applications and a solid set of test cases to go along with the change. I’m willing to suffer the small cloning performance penalty here in the name of maintainability and as a defense against memory leaks.

nateberkopec:
:+1: I think the only one that really makes a big difference is `newSnapshot.body.cloneNode(true)`. The rest aren't a big deal. I'll re-open with benchmarks - on a 4,000 element DOM it can save ~50ms or more.

PR #57, opened 2016-03-24
Make cloning a cache-level concern, speed up non-cacheable responses

nateberkopec:
This PR makes cloning the snapshots a concern of the Controller. 

Previously, cloning was performed as-needed in several locations. This PR clones all snapshots as they enter and leave the cache. This has a beneficial side-effect of removing unnecessary cloning during rendering when the Snapshot is not cacheable. 

[Before and after, tested on a 4,000-element DOM.](http://imgur.com/a/pl13b) If the response is not cacheable, this PR eliminates a costly ~70ms of `cloneNode` time when rendering the response.

In addition:

* delegate `isCacheable` to the Snapshot to make it similar to how Previews work
* `getSnapshot` no longer needs to know how to clone snapshots
* Reduce areas in the code that need to know about cloning
* Rename Cache to SnapshotCache, to more accurately reflect what it does.

### DIFF
--- a/src/turbolinks/controller.coffee
+++ b/src/turbolinks/controller.coffee
@@ -3,7 +3,7 @@
 #= require ./history
 #= require ./view
 #= require ./scroll_manager
-#= require ./cache
+#= require ./snapshot_cache
 #= require ./visit
 
 class Turbolinks.Controller
@@ -35,7 +35,7 @@ class Turbolinks.Controller
       @started = false
 
   clearCache: ->
-    @cache = new Turbolinks.Cache 10
+    @cache = new Turbolinks.SnapshotCache 10
 
   visit: (location, options = {}) ->
     location = Turbolinks.Location.wrap(location)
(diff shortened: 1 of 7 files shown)


src/turbolinks/cache.coffee
```diff
@@ -9,12 +9,12 @@ class Turbolinks.Cache
 
   get: (location) ->
     return unless @has(location)
-    snapshot = @read(location)
+    snapshot = @read(location).clone()
```
nateberkopec: After thinking about this some more, I'm not entirely sure *this* clone is necessary.

nateberkopec: [Aha, nvm, found your old justification here, which makes sense to me.](https://github.com/turbolinks/turbolinks-classic/pull/24#issuecomment-8944174)

nateberkopec:
Updated to fit the new rendering structure.

nateberkopec:
Updated again to reflect discussion at Railsconf - I moved `clone()` up a level to the Controller rather than the Cache. Also renamed Cache to SnapshotCache - the internal variables all refer to "snapshots", so I think it's appropriate to advertise that Turbolinks stores only Snapshots in that Cache.

sstephenson:
Excellent work Nate! Thanks for seeing this through. 👏

PR #61, opened 2016-03-29
Add a public method to change the URL in the browser location bar

jmbejar:
Just to provide some context, in our app we have a modal dialog containing a form. When the modal dialog is being shown we have a different URL (permalink) depending on if we are creating or editing an entry. However, if the user cancels or close the modal dialog, we want to just restore the previous URL (the one corresponding to the same page when the modal is hidden), but we also want to avoid a unnecessary call to `Turbolinks.visit` (because we are sure that nothing has changed in the page that contains the modal dialog).

This PR adds a way to change the URL without invoking `Turbolinks.visit`, so the subsequent request is not fired. This new function helps to change the URL in a Turbolinks-compatible way where this URL change will be part of the navigation history.

Being honest, I'm not sure if the addition of this function to the public API is worth :smiley: 
However, we are including this code in our own application, so I decided to share and see if it's interesting to others. Of course, it would be better for us if we can rely in a function that is part of the public API :relieved: 

What do you think? What would be a good name for this public function?

### DIFF
--- a/src/turbolinks/index.coffee
+++ b/src/turbolinks/index.coffee
@@ -12,3 +12,8 @@
 
   clearCache: ->
     Turbolinks.controller.clearCache()
+
+  changeURL: (location) ->
+    Turbolinks
+      .controller
+      .pushHistoryWithLocationAndRestorationIdentifier(location, Turbolinks.uuid())


Ribesg:
This is needed. Turbolinks prevents users from changing the current URL themselves, so Turbolinks should implement this ability themselves.

jmbejar:
@packagethief @sstephenson Have you had a chance to think in this proposal? I think it would be great to have a way to change the URL through the Turbolink API

nateberkopec:
@jmbejar Could you use URL fragments to implement the feature you're describing? 

Assuming Turbolinks can change how it works with URL fragments (#75) and ignore them completely, it seems to me you could implement your feature with URL fragments. Click the modal, add a fragment to the URL. When canceling the model, remove the fragment. You can write your own JS for opening the model if a user requests "yoururl#fragment". 

FWIW, as an aside, I don't think it's appropriate to use 2 different URLs to point to the same resource (your HTML page) in slightly different states. I think that's what fragments are for.

jmbejar:
I have a Rails application that just uses the default URL conventions given by Rails. For example, the URL for the page that renders a form is `/posts/new` (showing a model dialog in my case) and the URL to show this item is `/posts/:id`.

Following your idea, the post form would be shown when the URL is `/posts#new` or similar. However, I would miss the ability to automatically render the form page in the server and load it via Turbolinks because I'm not using the URL expected by the Rails app.
I could add JS logic triggered when the fragment is present to make an ajax call to `/posts/new` and get the HTML including the form, but it has a lot of overhead and it does not feel like how Rails and Turbolinks should work together.

Said that, I think the alternative of using URL fragments can make sense in other projects using Turbolinks outside the Rails world. It's an interesting idea that could help someone else there.

nateberkopec:
Hm, yeah, still not sure about this use-case. 

>(because we are sure that nothing has changed in the page that contains the modal dialog).

That's a very application-specific thing. Many applications cannot be sure that nothing has changed on a resource when navigating from `edit` to `new` (where a resource can be edited by many users, for example).

jmbejar:
I partially agree... this is a use case specific to our application, but it's something that can happen sometimes. Specially in sections or apps that allow to edit data that is not shared with other users (e.g. a personal board).

I'm proposing to make this option available as part of the public API because it does not sound like a very uncommon case to me, while it does not seem complicated to maintain in the long term (it's just a matter of exposing an internal method).

thiago-negri:
I have a search bar at top of some pages of my app and I would love to provide search-as-you-type feedback to the users, but if I do so (via remote form), I make it impossible for the user to share the search result with other people (no URL change).

I tried to trigger a visit with "search as you type", but then the input for search loses focus on page load. Giving a bad user experience if user wants to continue narrowing down the search.

Going the hash route seems to be much work, and also incur in fetching and rendering useless records for the given URL just to later be replaced by a JS fetching the search results for the query string on hash.

So, I would love to have a way to change "/products" to "/products?q=Ruby" with:
- Correct handling of Turbolinks history, i.e. Page state is saved and restored as needed;
- No "body" refresh, to avoid losing input focus;

I think it's something like "hey, Turbolinks, please consider this a new visit, I've already changed the page content, thanks."

Or even a "hey, Turbolinks, visit this new page but don't touch this element so I don't lose input focus" would work, I guess.

nateberkopec:
> I think it's something like "hey, Turbolinks, please consider this a new visit, I've already changed the page content, thanks."

Hm, yeah, there may be a broader story here around integrating with other SPA frameworks/SPA-like experiences.

vinc:
When I scroll down a page (http://example.com/photos), loading more content at the bottom via a "load more" button or infinite scrolling, I would like to have the URL reflecting the current page number (http://example.com/photos?page=42).

I think this is a common use case.

j-f1:
@packagethief @sstephenson What is the status on this?

ian:
@jmbejar any chance you'd roll this out as an plugin/extension so those who want to extend functionality can?

jmbejar:
@ian I don't feel like creating a separate library/plugin/extension for this functionality is a good move, because it would be just exposing an internal method, it's not adding too much in my opinion.  It is very little code to motivate the creation of a new library.

People who needs this behavior could just use the lines that I'm adding in this PR (and update them in the future if some internals change, this is a little risk).

pomartel:
I also have a use case for this pull request. When creating a new resource through a Rails remote form, the server returns a JSON response with the redirect path, and Turbolinks makes the redirection.

This all works fine except that when the user presses the browser's back button after the redirection, he's taken back to the new resource page (`/users/new`) instead of the edit page for the resource that was just created (`users/1/edit`). So instead of editing the created record, he will create a second one.

I get around this by replacing the current location in the history before visiting the redirect. Here's the client code I wrote to handle the remote form response:

```
$(document).on "ajax:success", ".remote-redirect-form", (e, data, status, xhr) ->
  Turbolinks.clearCache()
  location = xhr.getResponseHeader('Location')
  Turbolinks.controller.pushHistoryWithLocationAndRestorationIdentifier(location, Turbolinks.uuid())
  Turbolinks.visit(data.redirect)
```
The Location header is the edit path to the newly created resource and `data.redirect` is the path to the next step that Turbolinks visits. This fixes the back button issue for me.

m1guelpf:
Any update on this?

jspang:
This [answer on UX stack exchange](https://ux.stackexchange.com/questions/100866/should-opening-a-modal-box-change-the-browser-history) sums up nicely why this should be considered a must-have feature of Turbolinks.

At the moment I am doing a calendar with events popping up in a modal, which makes navigation far more intuitive than completely navigating away from the calendar. Each calendar event of course needs a unique URL change in the browser for people to copy and share, as well the ability to close the pop up modal (not navigate away) when the back button is hit.

This brillant oneliner PR seems to do the trick. @jmbejar - thank you!

akoury:
@nateberkopec I have a use case which would greatly benefit from this pull request.

Basically I have an ecommerce site, there I have products that have many variants that each have an identifier (think a shirt in red and the same shirt in blue), when I go to /products/3 I access a specific variant of a product (the red shirt), /products/4 could be a variant of the same product (the blue shirt).

In the site I have dropdowns that allow you to select from the different variants, if I am in the URI for the red shirt I can select the blue one and be taken to that site, however I do not wish to actually make the visit to that URI, since I have all the information from the variant already I can just update it in javascript so that it is instantaneous, however I would like to modify the URL so that it links to the blue shirt, that way if the customer decides to bookmark it he can return to that specific variant.

Thanks for taking the time to read this, this functionality would be really beneficial for some

robgibbons:
I would like to add my support for this feature request. Infinite scrolling when using Turbolinks is not able to update the URL which lets users bookmark and share page results. 

If Turbolinks is going to solely manage the history API then there should be support for some pass through usage via Turbolinks. Being able to update the URL without triggering a new visit is necessary for several use cases.

PR #65, opened 2016-03-31
Script evaluation

sstephenson:
This branch refactors the view to support `<script>` element evaluation.

The new rendering algorithm is as follows:

1. Ensure all `data-turbolinks-track="reload"` elements are the same in the current `<head>` and the new `<head>`.
   1. Find all tracked elements in the current `<head>`. Concatenate each element’s `outerHTML` to obtain the current tracked element signature.
   2. Find all tracked elements in the new `<head>`. Concatenate each element’s `outerHTML` to obtain the new tracked element signature.
   3. If the current and new tracked element signatures are not equal, stop rendering and reload the page.
2. Merge the new `<head>` into the current `<head>`.
   1. Find any `<style>` or `<link rel="stylesheet">` elements in the new `<head>`. Clone them and append them to the current `<head>.`
   2. Find any `<script>` elements in the new `<head>`. Activate them by manually cloning them (see below), then appending them to the current `<head>`.
   3. Find all provisional elements (everything except `<script>`, `<style>`, or `<link rel="stylesheet">`) in the current `<head>`, and remove them.
   4. Find all provisional elements in the new `<head>`. Clone them and append them to the current `<head>`.
3. Notify the application that Turbolinks will render the new `<body>`.
4. Replace the current `<body>` with the new `<body>`.
   1. Find any `<script>` elements in the new `<body>`. Activate them by manually cloning them, then replacing the original elements with the new elements.
   2. Iterate over all `data-turbolinks-permanent` elements in the new `<body>`. If any element has a corresponding permanent element with the same ID in the current `<body>`, replace the new element with the existing element.
   3. Assign the new `<body>` to `document.body`.
5. Notify Turbolinks.Visit to scroll the page as needed.
6. Focus the first element in `<body>` with the `autofocus` attribute.
7. Notify the application that Turbolinks has rendered.

You can trace this step-by-step by following the [`Turbolinks.SnapshotRenderer#render` method](https://github.com/turbolinks/turbolinks/blob/98108109f67f3983211e99022d8a560f0d2ea7b6/src/turbolinks/snapshot_renderer.coffee#L10-L18).

By default, browsers do not load or evaluate the content of `<script>` elements that were created by `innerHTML` or `cloneNode`. To activate a `<script>` element so it will be evaluated by the browser, we must clone it manually, first by creating a new, empty script element, then by copying over its attributes and text content.

In practice, this new behavior means:

1. Any script or stylesheet elements on the initial page load are evaluated by the browser, as you’d expect.
2. As you navigate from page to page, any new scripts or stylesheets in `<head>` will accumulate at the bottom in order. They’ll be evaluated once, as soon as they’re added, but not on subsequent page loads.
3. Any scripts in `<body>` will be evaluated every time Turbolinks renders a page. That means once for newly visited pages, once for pages restored from history cache, and twice for revisiting a page when the preview cache is enabled.

I think this is a reasonable default behavior, but we will probably need, at minimum, an annotation to tell Turbolinks not to evaluate a script.


MSchmidt:
Would this remove scripts/events when the user navigates away to a page that doesn't contain the same script tag? E.g. I'd use this to have certain scripts only execute on matching pages so my javascript is as slim as possible on each page. I wouldn't want new script tags to be added over and over again.

sstephenson:
> Would this remove scripts/events when the user navigates away to a page that doesn't contain the same script tag? E.g. I'd use this to have certain scripts only execute on matching pages so my javascript is as slim as possible on each page. I wouldn't want new script tags to be added over and over again.

See the list at the bottom of the original message. It depends on whether your scripts are in `<head>` or `<body>`. In the `<head>`, new scripts accumulate with old scripts as you navigate. In `<body>`, Turbolinks evaluates scripts every time it renders a new page.

MSchmidt:
Ah sorry - maybe I wasn't very clear what my question actually was. I understand when new scripts evaluate but would they also de-evaluate at some point?

Let's say I have a script in ``<head>`` that adds an ``EventListener`` on ``turbolinks:load``. Then I navigate away to a page that doesn't have that script. So the script would disappear from ``<head>`` but would the EventListener go away? Or would it stick around and execute on ``turbolinks:load``?

sstephenson:
No, there’s no such thing as “de-evaluation”—you’ll have to handle cleanup yourself.

nateberkopec:
>  twice for revisiting a page when the preview cache is enabled. 

Hmm, this seems like it will trip people up, even though it makes sense from Turbolinks' perspective.

src/turbolinks/view.coffee
```diff
@@ -29,58 +31,8 @@ class Turbolinks.View
...
-    if script.hasAttribute("src")
-      element.src = script.getAttribute("src")
-    else
-      element.textContent = script.textContent
-    element
-
-  focusFirstAutofocusableElement = ->
-    document.body.querySelector("[autofocus]")?.focus()
+  renderError: (error, callback) ->
+    Turbolinks.ErrorRenderer.render(@delegate, callback, html)
```
packagethief: Typo? `html` -> `error`

sstephenson: 👼 f145ca2

sstephenson:
> Hmm, this seems like it will trip people up, even though it makes sense from Turbolinks' perspective.

I think it’s the sort of thing that’s going to trip people up no matter what we do.

The main cases we want to support are:

1. Loading new `<script src>` elements in `<head>` during Turbolinks navigation. If a page you navigate to depends on a script that wasn’t on the initial page, we ought to load it.
2. Loading contextual, page-specific data in inline `<script>` elements in `<body>`.

Any other operations, like installing event listeners or setting up other behavior, should live in event handlers in your application’s JS bundle.

sstephenson:
1ce1876 adds support for a `data-turbolinks-eval="false"` annotation on script elements.

trevorturk:
👋 I ran into a problem where I'm unable to see embedded GitHub gists after visiting a page using Turbolinks. I made a simple Rails app to demonstrate the issue here: https://github.com/trevorturk/turbogist

I came across a workaround here: http://josephndungu.com/tutorials/using-github-embedded-gists-with-turbolinks but I decided to just disable Turbo links for pages with embedded gists as described here: https://github.com/turbolinks/turbolinks#disabling-turbolinks-on-specific-links because it seemed like a bit much. 

Apologies in advance if I'm missing something obvious, but I thought even if this is an known issue that we can't work around (eval'ing javascript from a different domain?) it would be worth documenting here for future googlers. 

Thank you! We're loving Turbolinks 5 at @IFTTT -- it was an easy upgrade and gave us a noticeable speed improvement over the previous version. 💝

javan:
> but I thought even if this is an known issue that we can't work around (eval'ing javascript from a different domain?)

@trevorturk, that's not the issue here. Gist embed scripts (attempt to) do a bunch of `document.write`s, which won't work for dynamically added content. You may have noticed console warnings like this when TL-navigating to a page with gist scripts.

```
Failed to execute 'write' on 'Document': It isn't possible to write into a document from an asynchronously-loaded external script unless it is explicitly opened.
```

Here's one way you might handle them. Similar to the workaround you posted above, but optimized for Turbolinks 5 events:

```coffee
scriptSelector = "script[src*='gist.github.com']"
placeholderSelector = "[data-gist-src]:not([data-gist-loaded])"

addEventListener "turbolinks:before-render", (event) ->
  for script in event.data.newBody.querySelectorAll(scriptSelector)
    placeholder = document.createElement("div")
    placeholder.dataset.gistSrc = script.src
    script.parentNode.replaceChild(placeholder, script)

addEventListener "turbolinks:render", ->
  for placeholder in document.body.querySelectorAll(placeholderSelector)
    loadGist(placeholder)

loadGist = (placeholder) ->
  placeholder.dataset.gistLoaded = true
  url = placeholder.dataset.gistSrc.replace(".js", ".json")
  $.ajax(url, dataType: "jsonp").then (gist) ->
    createStylesheet(gist.stylesheet)
    placeholder.innerHTML = gist.div

createStylesheet = (href) ->
  unless document.head.querySelector("link[rel=stylesheet][href='#{href}']")
    link = document.createElement("link")
    link.rel = "stylesheet"
    link.href = href
    document.head.appendChild(link)
```

Issue #74, opened 2016-04-15
Turbolinks.visit(url, action: 'back')


glennfu:
Imagine a crud situation where you're on the 'show' action for an item. You click 'edit'. You type in some changes, hit 'Save'. The server redirects you to the 'show' action. Normally in a web situation this is OK. The history stack is then 'somewhere', 'show', 'edit', show'. However, when building an iOS app by hooking up to turbolinks-ios, a different convention is expected. In that situation it's more common that the 'edit' comes in as a modal (maybe), you do your edits, then the edit form is 'popped' and you're dropped back on the 'show' view with the updated content. At any rate, you expect that once you're dropped back to the 'show', that pressing 'back' would send you to the 'somewhere' page that you came from originally, not to the edit form, and then back to the 'show' yet again.

Using `Turbolinks.visit(url, action: 'replace')` gives us two 'show' actions in a row on the stack. That means you see the results, click 'back' and you see the same 'show' action again. Still not desirable.

Now you could conceivably track the referrer when the 'edit' view is loaded, and if it was the 'show' view, you could, add something to the submission to let the server know that instead of sending `Turbolinks.visit()`, it should call `window.history.back()`. However this seems like a big brittle hack.

Instead I think a better solution would be to call `Turbolinks.visit(url, action: 'back')`. The expected behavior would be that if the previous url in the history is the url provided, then pop to that url, refreshing it. Otherwise, visit it like normal. I think this would give the best experience here.

What do you all think? Is this a reasonable idea? Or too much of a non-standard edge case to live as part of Turbolinks?

gregblass:
+1, came here looking to accomplish this. My case: Index page, click through to a show. Normally you'd want the back button to go back to the index page.

But what if they're coming from an activity feed? If they click back, it should in this case send them back to the activity feed.

glennfu:
It also just occurred to me that, in iOS, the 'popToRootViewControllerAnimated' method exists for times when you have a deeply nested navigation stack and you want to reset to the 'home', either from a general 'home' link or maybe a Sign Out button that resets you to root_path. In either case, you'd expect there not to be a 'Back' button to take you to the left any further. Would this justify a 'reset' action as well?

gregblass:
+1, I came across the same problem

elsurudo:
Cross-posting a workaround from the other ticket, in case it may help anyone: https://github.com/turbolinks/turbolinks/issues/88#issuecomment-263567672

rosendi:
[...]

PR #76, opened 2016-04-15
Support action: 'back' to navigate back in history

glennfu:
I decided to take a stab at this to see how it would work after posting #74. There's still an issue remaining where in the case of calling `Turbolinks.visit("http://site.dev/path/here", {"action":"back"})` Firebug reports that `http://site.dev/path/here` is called once, aborted, then called again. I'm having trouble figuring out why that is.

On the turbolinks-ios side, in my `ApplicationController.swift#presentVisitableForSession` I had to add `else if action == .Back { popViewControllerAnimated(true) }` and in `Turbolinks/Actions.swift` I added `case Back    = "back"`.

With those changes, everything seemed to work perfectly! So aside from the unexplained aborted request, this does exactly what I hoped it would do. I look forward to hearing what you think! If I'm on the right track, let me know and I'll try to put in some appropriate tests and another pull request to turbolinks-ios. I haven't yet gotten far with turbolinks-android but I hope to dive into that soon also.


glennfu:
If anyone's trying to implement this on their own in the meantime, the https://github.com/turbolinks/turbolinks-rails gem needs `redirection.rb#26` changed to `action: ["advance", "replace", "back"].include?(action.to_s) ? action : "replace"`

gregblass:
So just wondering how I would go about using this. I'm trying:

`gem 'turbolinks', git: 'https://github.com/glennfu/turbolinks.git', branch: "navBack"`

But thats not working. Then I'd also have to modify turbolinks-rails redirection.rb#26 as well?

nateberkopec:
@gregblass I actually just clone down the repo and run `blade build`. I then copy `dist/turbolinks.js` into the `vendor/assets/javascripts` folder of my app. This makes it easier to hack on/modify while you're investigating the changes.

Fudoshiki:
any tmp solution?

glennfu:
@Fudoshiki My Rails app has `gem 'turbolinks', github: 'glennfu/turbolinks-rails'` in the Gemfile. Then I manually include this in my application.js instead of the original turbolinks.js: https://gist.github.com/glennfu/34bbe7e7b1dc5d4cfeb84c82f423658d

I've been running this for 2 months now with no problems.

Fudoshiki:
I have next scenario:

REST
1. Open form with method: :put (edit action)
2. Enter invalid data
3. Submit
4. Got rendered page with errors (update action)
5. Press F5 request GET on update action (route error)

Same issue? 

gem 'turbolinks', github: 'glennfu/turbolinks-rails' and https://gist.github.com/glennfu/34bbe7e7b1dc5d4cfeb84c82f423658d don't solved this

glennfu:
I don't think there's ever a proper setup where F5 would resubmit a form. It's certainly not desirable behavior. At any rate, unless I'm misunderstanding your intent, I don't know what this has to do with this Issue, which is about adding the ability to navigate backward in the stack.

In your described scenario, navigating back wouldn't even be the expected behavior. More likely you would, in your controller, tell Turbolinks to redirect with `action: 'back'` on a successful form submit, then the browser would drop you on the page you were on BEFORE you saw the form.

I believe what you are seeing on rendering page errors should be `action: 'replace'` in order to update the same page you're on, which is already part of Turbolinks without this PR.

glennfu:
Just realized one more important change. Line 215 of `Session.swift` in turbolinks-ios reads `visitVisitable(visitable, action: .Restore)`. This should be changed to `visitVisitable(visitable, action: .Back)`

glennfu:
I just pushed a commit that fixes a bug I discussed here: https://github.com/turbolinks/turbolinks-ios/issues/50

src/turbolinks/history.coffee
```diff
@@ -20,16 +23,27 @@ class Turbolinks.History
...
   push: (location, restorationIdentifier) ->
     location = Turbolinks.Location.wrap(location)
+    locationStack.push(location)
     @update("push", location, restorationIdentifier)
 
   replace: (location, restorationIdentifier) ->
     location = Turbolinks.Location.wrap(location)
     @update("replace", location, restorationIdentifier)
 
+  back: (location, restorationIdentifier) ->
```
nateberkopec: Turbolinks actions are verbs. `back` is generally not used as a verb, so perhaps another word could be used for this action. `return`?

src/turbolinks/history.coffee
```diff
@@ -20,16 +23,27 @@ class Turbolinks.History
...
+    location = Turbolinks.Location.wrap(location)
+    
+    previousUrl = locationStack[locationStack.length - 2]
+    if previousUrl?.absoluteURL == location.absoluteURL
+      window.history.go(-1)
```
nateberkopec: can just use `history` here.

src/turbolinks/history.coffee
nateberkopec: Actually, given the actual behavior of this PR, and your comment that "The expected behavior would be that if the previous url in the history is the url provided", perhaps `refresh` is a better name.

src/turbolinks/history.coffee
```diff
@@ -20,16 +23,27 @@ class Turbolinks.History
...
 
   replace: (location, restorationIdentifier) ->
     location = Turbolinks.Location.wrap(location)
     @update("replace", location, restorationIdentifier)
 
+  back: (location, restorationIdentifier) ->
+    location = Turbolinks.Location.wrap(location)
+    
+    previousUrl = locationStack[locationStack.length - 2]
+    if previousUrl?.absoluteURL == location.absoluteURL
```
nateberkopec: Man, what a total bummer that you can't access the entries of the History API. Security, always getting in the way ;)

src/turbolinks/history.coffee
glennfu: I feel like `refresh` sounds synonymous to `reload` as in "refresh the page", which would suggest history doesn't move. In this case, history is indeed going "back". Other synonyms of `go_back`, `retire`, `retreat`, none of those sound good! I think `return` makes the most sense, but it's also a bit confusing. If you were instructing someone to go to the previous page, how would you say it? "Hit back" or "Go back". I guess you could say "Return to the previous page" but you could just as easily say "Return to the home screen". "Return" by itself doesn't accurately describe it as well as "go back". Hmmm!

There's my two cents. You can pick whatever name you think is best and I'll update it :)

src/turbolinks/history.coffee
glennfu: Agreed! I hate that this feels like a hack, but it seems to work pretty solidly so far.

src/turbolinks/history.coffee
nateberkopec: Yeah - just to match how it's called in `update`.

glennfu:
Hey @nateberkopec, any opinion on my last message about "back" vs "go_back" vs "refresh"? I'm less excited about "refresh" because this PR's behavior is about simulating the "back" button, and "refresh" is synonymous with "reload". Also, gotta throw in there that Webster says "back" is a verb too :)

Let me know what word we should use here and I'll update this PR.

nateberkopec:
@glennfu Probably just `go_back`

"to back", back's verb form, as in "to back a candidate" doesn't really fit here IMO.

nateberkopec:
but that's just my opinion, the maintainers of this repo I know usually have really strong opinions on readability/naming.

It's also possible that they're thinking up some different solution entirely for this problem (though your PR still helps by proposing one).

michaelphines:
`retreat`?

glennfu:
I just noticed the only other context this "verb" is used in the code is in the `backHistoryWithLocationAndRestorationIdentifier` which is analogous to the `replaceHistoryWithLocationAndRestorationIdentifier` and `pushHistoryWithLocationAndRestorationIdentifier` methods. With that in mine, `rewindHistory` and `reverseHistory` sound pretty good. I don't know that `goBackHistory` sounds good. `retreatHistory` is a solid "maybe" :)

@sstephenson @gregblass @packagethief any opinions on this verb? I'm happy to update the entire working and rebased solution once I feel like we generally have a good name for it. It would be awesome if we can get it right the first time because this verb will exist in turbolinks, turbolinks-ios, turbolinks-android, and turbolinks-rails, so refactoring it later if we decide we don't like it will be really annoying.

WaKeMaTTa:
`backward` :question: 

 (`backwardHistory` & `backwardHistoryWithLocationAndRestorationIdentifier`)

Startouf:
What about `rewind` ?

michaelphines:
Has anyone suggested `popHistory` yet?

dgilperez:
+1 to `rewind`. `go_back` or `step_back` would be also OK IMO.

Fudoshiki:
any progress?

domchristie:
> With that in mind, rewindHistory and reverseHistory sound pretty good. I don't know that goBackHistory sounds good. retreatHistory is a solid "maybe" :) … any opinions on this verb?

@glennfu I don't feel that `back` is that bad. It may not be consistent with the other verbs, but it is clear and straightforward. For me, the issue with `rewind` and `reverse` is that they sound like they could transport the user to _any_ point in a continuous timeline (like rewinding a cassette, or reversing a car), rather than traveling to the last discrete step.

However, with this in mind, I wonder if this approach could be expanded to navigate to any location in the stack, rather than just the previous one? For example, given the following stack:

```
[Index] [Show] [Edit]
```
then calling `Turbolinks.visit('Index', { action: 'back' })` could result in the following stack:

```
[Index]
```

This could be a solution to the "reset" case mentioned here: https://github.com/turbolinks/turbolinks/issues/74#issuecomment-213499482, but I can't work out if this might cause confusion in some other cases.

src/turbolinks/history.coffee
domchristie: `Turbolinks.Location` has a handy [`isEqualTo`](https://github.com/turbolinks/turbolinks/blob/c340c42e2ece8c057aec611074971462a6e76c71/src/turbolinks/location.coffee#L43-L44) method which handles this check, so this could be update to:
```suggestion
    if location.isEqualTo(previousUrl)
```

and perhaps `previousUrl` could be renamed to `previousLocation`?

Issue #89, opened 2016-05-05
Does not degrade gracefully on IE9


fschwahn:
When loading a page which includes turbolinks in IE9, Turbolinks causes an error. I think the reason is that IE9 does not implement `classList` - however adding the polyfill from https://developer.mozilla.org/en/docs/Web/API/Element/classList didn't make the error go away. If I remove turbolinks completely, the javascript error goes away.

I'd expect Turbolinks not to cause an error, even if included in a browser which is not supported.

jonchay:
👍 on this issue. We have an error raised:
```
SCRIPT5007: Unable to get value of the property 'add': object is null or undefined
```

nateberkopec:
Agreed, can you guys check out #90 and see if that fixes the issue for you?

PR #90, opened 2016-05-05
Only initializing Turbolinks if supported.

bradly:
Hi all,

We are getting in error in IE9 that we think is do to the progress bar trying to be created even when Turbolinks.supported is false. I believe this is the issue reported here: #89.

I don't really know what I'm doing here, so let me know if this isn't the correct approach.

Thanks!


packagethief:
Thanks for the PR, @bradly. We still want to expose `Turbolinks` in unsupported browsers so your application doesn't need to worry about support when using `Turbolinks.visit`. In unsupported browsers, `Turbolinks.visit` degrades to setting `window.location`.

Resolved in 75a6a55 by avoiding `Element.classList` altogether. Unnecessary, since the progress bar only uses a single class name.

bradly:
@packagethief Awesome, glad it's resolved :) <3

PR #94, opened 2016-05-12
Use getElementById instead of querySelector

nanaya:
I noticed that specifying invalid id string as hash breaks turbolinks. My guess says it also happens when restoring (going back) to a page with invalid hash (which possibly user-specified).

Using `getElementById` allows any string to be used (or is it?).

    > document.getElementById('123')
    null
    > document.querySelector('#123')
    SyntaxError: An invalid or illegal string was specified

From what I can tell `@location.anchor` always come from slicing `location.hash`.


nanaya:
Whoops nvm.

PR #96, opened 2016-05-12
Prevent non-id hash from breaking navigation 

### LINKED ISSUE #95: Non-id hash breaks navigation
1. visit a turbolink page with non-id hash (`http://somewhere.com/a#123`)
2. navigate within turbolinks scope
3. go back
4. `SyntaxError: An invalid or illegal string was specified` is raised when calling `snapshot.coffee`'s `hasAnchor`

I tried fixing it by replacing with `getElementById` but noticed the variable doesn't have the function so I replaced it by [prepending the call with `try`](/turbolinks/turbolinks/compare/master...nanaya:any-hash) and it seems to work. I can't reopen the pull request though (probably because it's force-pushed?) and since there's no example of `try` usage in codebase I'm opening this issue instead (and maybe someone else have better idea).

nanaya:
Attempt 2 to fix #95.

### DIFF
--- a/src/turbolinks/snapshot.coffee
+++ b/src/turbolinks/snapshot.coffee
@@ -27,7 +27,7 @@ class Turbolinks.Snapshot
     @getSetting("cache-control")
 
   hasAnchor: (anchor) ->
-    @body.querySelector("##{anchor}")?
+    try @body.querySelector("##{anchor}")?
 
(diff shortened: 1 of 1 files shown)


nanaya:
Sorry, the commit message is terrible (I was in the middle of doing something else).

PR #101, opened 2016-05-17
Add UMD support. Fixes #91

### LINKED ISSUE #91: "Uncaught ReferenceError: Turbolinks is not defined" when required with Browserify
We are using Browserify to build our front-end assets. We successfully installed the `turbolinks` npm package and compiled the assets, but when the page loads in the browser, `Uncaught ReferenceError: Turbolinks is not defined` is raised in the package code `dist/turbolinks.js` at `Turbolinks.copyObject=function(t){var e,r,n;r={}; ...`

Oddly, we are able to use the package if we paste the raw `dist/turbolinks.js` into a script tag in the html head.

Are there any steps we need to take in order to get this package to work with Browserify? If Browserify is currently unsupported, are you open to pull requests to introduce support? Any pointers would be appreciated.

Thanks

javan:
Define UMD module for Turbolinks to support script / module build tools like webpack, browserify, require.js, etc.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,5 +7,5 @@ gem 'coffee-script-source', '~> 1.9.1'
 gem 'eco'
 gem 'uglifier'
 
-gem 'blade', github: 'javan/blade'
+gem 'blade', '~> 0.5.2'
 gem 'blade-sauce_labs_plugin', github: 'javan/blade-sauce_labs_plugin'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,33 +8,29 @@ GIT
       faraday
       selenium-webdriver
 
-GIT
-  remote: git://github.com/javan/blade.git
-  revision: 84e843ed5517c8eff61d876dcf4634deebe88aa8
-  specs:
-    blade (0.5.0)
-      activesupport (>= 3.0.0)
-      blade-qunit_adapter (~> 1.20.0)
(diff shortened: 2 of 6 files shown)


sstephenson:
We’ll need to update [the documentation](https://github.com/turbolinks/turbolinks#installation-using-webpack) too.

javan:
What do you think about removing the [Installation Using Webpack](https://github.com/turbolinks/turbolinks#installation-using-webpack) documentation since it Just Works now?

sstephenson:
I’m fine with removing the Webpack-specific parts, but we should still have a parallel set of instructions for using Turbolinks as a module. You still have to install the package and require it somewhere.

javan:
Updated the docs

javan:
FYI, I confirmed the compiled .js works normally with `<script src="turbolinks.js"></script>` and with webpack and browserify:
```js
(function() {
  var Turbolinks = require("./turbolinks");
  console.log("local var Turbolinks =", Turbolinks); // defined
})()

addEventListener("DOMContentLoaded", function() {
  console.log("widow.Turbolinks =", window.Turbolinks) // undefined
})
```

TSMMark:
As a Browserify user, I thank you! :+1:

sstephenson:
Thanks for making this possible @javan! 💝🎉

PR #123, opened 2016-06-22
Use an HTTPs git link for a gem

AnthonySuper:
Using the `github:` option in a Gemfile is vulnerable to MITM attacks according to [bundler's documentation](http://bundler.io/git.html). This change uses the `git:` option with an HTTPs URL instead, which is not vulnerable.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,4 +8,4 @@ gem 'eco'
 gem 'uglifier'
 
 gem 'blade', '~> 0.5.2'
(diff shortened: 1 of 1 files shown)


sstephenson:
Thanks for the patch! We’re actually not testing on Sauce Labs at the moment so I just [removed that gem altogether](https://github.com/turbolinks/turbolinks/commit/af29e2100677ff70436973014644815063dc2664).

PR #131, opened 2016-07-01
Remove reference to "beta" version

tom-lord:
Although the old command was valid, it's now slightly misleading to display this as "in beta", in the README!


javan:
Thanks!

PR #141, opened 2016-07-14
Fix behavior of data-turbolinks-preview attribute

darfux:
I'm using AngularJS with Rails5 and I found the cache of turbolinks may cause angular's re-bootstrap problem when click the back button of browser. 

I tried to solve it with the _data-turbolinks-preview_ attribute but it seems that the attribute doesn't work as it was described at [Detecting When a Preview is Visible](https://github.com/turbolinks/turbolinks#detecting-when-a-preview-is-visible). So I fixed it and added a testcase:)

### DIFF
--- a/src/turbolinks/visit.coffee
+++ b/src/turbolinks/visit.coffee
@@ -56,7 +56,7 @@ class Turbolinks.Visit
 
   loadCachedSnapshot: ->
     if snapshot = @getCachedSnapshot()
-      isPreview = @shouldIssueRequest()
+      isPreview = not @shouldIssueRequest()
       @render ->
         @cacheSnapshot()
         @controller.render({snapshot, isPreview}, @performScroll)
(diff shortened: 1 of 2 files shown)


sstephenson:
I’m afraid I don’t understand... the `data-turbolinks-preview` attribute is working exactly as designed. A loaded snapshot is a _preview_ if it’s going to be replaced by a subsequent fresh page load. So application visits to cached pages will show previews, but restoration visits will not.

darfux:
Thanks for reply! I'm sorry for misunderstanding the document... But it means that there's no way to determine whether a page is loaded from the **network** or **cache**? In my case, the AngularJS rendered the original html before turbolinks caches it. So I need to know this to avoid re-rendering on `turbolinks:load`event when the page is loaded from cache, as it has been rendered.

Issue #150, opened 2016-07-26
Should allow specifying history object to use (not always window.history)


TSMMark:
### Why I want this feature:

I can not get turbolinks to play nice with react-router. After a turbolinks navigate, the router still continues to render the previous route, as if the browser history has not changed. I believe this is due to react-router using a custom history object `browserHistory` which is from [`history`](https://www.npmjs.com/package/history). I think it is not able to listen to `window.history` changes, which is what turbolinks uses (see here: https://github.com/turbolinks/turbolinks/blob/master/src/turbolinks/history.coffee#L46)

### Proposed solution:

Instead of always using `window.history`, we may be able to pass in `history` as an option to Turbolinks which it can make calls to `pushState` and `replaceState`.

PR #151, opened 2016-07-27
Check for window.turbolinksHistory first and then use window.history if not exist

TSMMark:
Corresponding Issue: #150


sstephenson:
Thanks for the pull request. Can you please point me to a page demonstrating the problem this solves? I’m skeptical that Turbolinks and react-router will ever be compatible.

sstephenson:
Also, let’s please avoid hostile or aggressive [commit messages](https://github.com/turbolinks/turbolinks/pull/151/commits/1a8c4e031c1100f6a103cd92e10d411828510c72). Thanks.

TSMMark:
I won't be able to include a link to a page, but I can post our relevant js:

```js
var _ = require("lodash");
var ReactRouter = require("react-router");

var removeDomainAndProtocol = function(url) {
  return url.replace(/^.*\/\/[^\/]+/, "");
}

// API Adapter to handle turbolinks & react-router.
var turbolinksHistory = _.extend({}, ReactRouter.browserHistory);

turbolinksHistory.pushState = function(_turbolinks, _null, urls) {
  turbolinksHistory.push(removeDomainAndProtocol(urls.requestURL));
}
turbolinksHistory.replaceState = function(_turbolinks, _null, urls) {
  turbolinksHistory.replace(removeDomainAndProtocol(urls.requestURL));
}

module.exports = window.turbolinksHistory = turbolinksHistory;
```

```jsx
// Router uses turbolinksHistory
var turbolinksHistory = require("./turbolinksHistory");
<Router history={turbolinksHistory}>
  {Routes}
</Router>
```

TSMMark:
Bugs were discovered with the approach. We did get react-router and turbolinks functioning together though. For the record, future developer who has stumbled upon this, you have to unmount and then re-mount your react-router on every turbolinks navigation. You can not use react-router `Link` components, etc, nor can you use any react-router navigation functions.

You also must create a new history object on every page load, kind of like this:

```jsx
// On every turbolinks:before-visit
var onTurbolinksPageVisit = function () {
  var createHistory = require("history").createHistory;
  // We have to create a new history every page navigation because of turbolinks.
  // You see, turbolinks makes calls to the window.history directly, and react-router
  // history has no way of knowing about it. If you don't create a new history object
  // every time, the history object in memory will contain cached data that makes
  // react-router believe that it is still on the previous page.
  var history = useRouterHistory(createHistory)({});
  var routerComponent = (
    <Router history={history}>
      {Routes}
    </Router>
  );
  // ... Handle the mounting of routerComponent
}
```

PR #166, opened 2016-08-11
Fire "turbolinks:load" event even for unsupported browsers [Fixes #130]

### LINKED ISSUE #130: Turbolinks not emitting events for "unsupported" browsers
Perhaps this is "feature not a bug" territory, but _seems_ like before https://github.com/turbolinks/turbolinks/commit/14f07ff12343f77f393ac6c7b571322f3b4e493f, events like `turbolinks:load` and friends would still be fired, which was useful as it meant that all the other JS would still continue to work, even if turbolinks wasn't supported (I'm possible completely wrong here...)

balvig:
This is an attempt to fix the issue mentioned in #130, where `turbolinks:load` doesn't get fired for unsupported browsers.

Not sure if this is the best way to go about it, but this PR moves the `Turbolinks.supported` conditional to `history.replace()` (which is where things fall apart when run on an unsupported browser) so that the initialization process which fires the event in question still runs its course.

I've added a test as well to assert that the event gets fired on a browser without `windows.pushState`. Forgive my ignorance since this is my first time to use blade, but I'm not sure if I'm doing this the right way either! 

I appreciate any pointers in case this is completely off track! 🙇

### DIFF
--- a/src/turbolinks/controller.coffee
+++ b/src/turbolinks/controller.coffee
@@ -15,7 +15,7 @@ class Turbolinks.Controller
     @clearCache()
 
   start: ->
-    if Turbolinks.supported and not @started
+    unless @started
       addEventListener("click", @clickCaptured, true)
       addEventListener("DOMContentLoaded", @pageLoaded, false)
       @scrollManager.start()
--- a/src/turbolinks/history.coffee
+++ b/src/turbolinks/history.coffee
@@ -23,8 +23,9 @@ class Turbolinks.History
     @update("push", location, restorationIdentifier)
 
   replace: (location, restorationIdentifier) ->
-    location = Turbolinks.Location.wrap(location)
-    @update("replace", location, restorationIdentifier)
+    if Turbolinks.supported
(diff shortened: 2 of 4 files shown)


balvig:
@sstephenson I'm guessing you just need to get around to it, but in case this PR itself has some issues, let me know if there is anything I can do to make it easier to review!

balvig:
For anyone having the same problem, here is the workaround we currently use:

```coffee
# Wrapper for `turbolinks:load`, so it works with IE9 until
# https://github.com/turbolinks/turbolinks/pull/166 gets merged
if Turbolinks.supported
  $(document).on "turbolinks:load", -> $(this).trigger("page:show")
else
  $(document).on "DOMContentLoaded", -> $(this).trigger("page:show")
```

then we only hook into that custom `page:show` event and avoid the `turbolinks:load` event, ie:

```coffee
$(document).on "page:show", ->
  alert "Show on every page whether the browser supports turbolinks or not"

balvig:
(cleaning house in my list of stale pull requests, closing this for now 🙇)

Issue #173, opened 2016-09-02
Turbolinks triggers exception on IE8


jonwolfe:
We've been relying on the statement that "In unsupported browsers, Turbolinks gracefully degrades to standard navigation," but unfortunately that's not the case with IE8 -- it's triggering javascript exceptions.

I don't expect TL to do anything on such an old browser, of course. That's not the issue. But if you make the claim that it gracefully degrades, please verify the claim so that we don't make incorrect assumptions.

<img width="693" alt="screen shot 2016-09-02 at 10 10 43 am" src="https://cloud.githubusercontent.com/assets/5140/18212160/4e31c692-70f6-11e6-96d9-cd5873a3fbe4.png">

javan:
Thanks! Got a fix ready in #174 that we'll get into the next release. In the meantime you could add a polyfill for `addEventListener` (here's one https://gist.github.com/jonathantneal/3748027). It won't enable Turbolinks, but it should shush that error.

jonwolfe:
Thank you, Javan!

PR #174, opened 2016-09-03
Avoid error calling addEventListener in browsers that don't support it

javan:
Fix for #173

### DIFF
--- a/src/turbolinks/history.coffee
+++ b/src/turbolinks/history.coffee
@@ -1,21 +1,16 @@
-pageLoaded = false
-
-addEventListener "load", ->
-  Turbolinks.defer ->
-    pageLoaded = true
-, false
-
 class Turbolinks.History
   constructor: (@delegate) ->
 
   start: ->
     unless @started
       addEventListener("popstate", @onPopState, false)
+      addEventListener("load", @onPageLoad, false)
       @started = true
 
   stop: ->
     if @started
       removeEventListener("popstate", @onPopState, false)
+      removeEventListener("load", @onPageLoad, false)
       @started = false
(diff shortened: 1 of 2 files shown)


src/turbolinks/index.coffee
```diff
@@ -7,7 +7,9 @@
 
 @Turbolinks =
   supported: do ->
-    window.history.pushState? and window.requestAnimationFrame?
+    window.history.pushState? and
+      window.requestAnimationFrame? and
+      window.addEventListener?
```
javan: Not technically needed to fix the issue, but seems reasonable to add for clarity.

src/turbolinks/history.coffee
```diff
@@ -35,11 +30,18 @@ class Turbolinks.History
...
+
   # Private
 
   shouldHandlePopState: ->
     # Safari dispatches a popstate event after window's load event, ignore it
-    pageLoaded is true
+    @pageIsLoaded()
+
+  pageIsLoaded: ->
+    @pageLoaded or document.readyState is "complete"
```
javan: Added an additional check for `document.readyState` to cover the theoretical case where turbolinks is loaded async after the `load` event.

nateberkopec: 👍 Similar to [what JQuery does](https://github.com/jquery/jquery/blob/master/src/core/ready.js#L80).

printercu:
Is there plan to release this fix? I've found that `addEventListener("load", ...)` is not firing event sometimes in iOS simulator, so `history.back()` doesn't work. This patch seems to fix the issue.

Issue #75, opened 2016-04-15
Turbolinks should follow same-page anchor links without reloading the page


preetpalS:
Consider the following HTML link to an element with an id of `bookmark`:

`<a href="#bookmark" >Bookmark within page</a>`

Turbolinks currently intercepts this link and prevents the browser from following it unless you add an annotation on the link (`data-turbolinks="false"`). Once Turbolinks has intercepted the link, it makes another HTTP request for the same location again (but with the id appended at the end (`/url#bookmark`)) instead of just scrolling to the anchored element. Since bookmark links within the same page normally do not result in an HTTP request, one could argue that the default behavior of Turbolinks for links that have an `href` attribute beginning with `#` results in worse performance (due to the network request) and violates the principle of least surprise.

Is it possible to disable Turbolinks for all links that have an `href` attribute beginning with `#` without having to manually annotate them with `data-turbolinks="false"`?

I am currently using turbolinks (5.0.0.beta2) (turbolinks-source (5.0.0.beta4)) with Rails 4.2.6.

packagethief:
> Since bookmark links within the same page normally do not result in an HTTP request, one could argue that the default behavior of Turbolinks for links that have an href attribute beginning with # results in worse performance (due to the network request) and violates the principle of least surprise.

I agree. We should ignore clicks on links that are anchors to the current page, and let them fall through to the browser.

preetpalS:
Here is a workaround to prevent Turbolinks from making unnecessary network calls:

```
document.addEventListener("turbolinks:before-visit", function(event) {
    var origin = window.location.href; // Current URL
    var destination = event.data.url; // Destination URL

    // Make sure both destination and origin urls end with "/" unless they contain an "#"
    if ((origin.match(/#/) == null) && origin[origin.length - 1] !== "/") origin = origin + '/';
    if ((destination.match(/#/) == null) && destination[destination.length - 1] !== "/") destination = destination + '/';
    var shorterLength = Math.min(origin.length, destination.length);

    // Prevent Turbolinks from doing anything when clicking on a link to the current page
    if (origin === destination) {
        event.preventDefault();
        return;
    }

    // See if the first `shorterLength` characters of both the origin and destination URLS are the same.
    // If these characters are not the same, do nothing.
    if (origin.substring(0, shorterLength) !== destination.substring(0, shorterLength)) {
        return;
    }

    // If the origin URL contains an "#" immediately after the first `shorterLength` characters, do nothing.
[...]

PR #175, opened 2016-09-07
Add support for scrolling to named anchors

domchristie:
This pull request adds support for scrolling to named anchors, so that calling `Turbolinks.visit('/path#name')`, when the target page includes `<a name="name"></a>`, behaves the same as if the page included an element with an ID of `name`.

The current behaviour ignores named anchors as demonstrated [here](https://powerful-wildwood-35478.herokuapp.com).

I understand that using anchors in this way is rather old fashioned (indeed it is obsolete in HTML5), but visiting `/path#name` directly versus via Turbolinks results in different behaviour, so I thought it'd be worth fixing.

I'd be happy to add tests if you think it is necessary, but I may need a pointer or two.

### DIFF
--- a/src/turbolinks/controller.coffee
+++ b/src/turbolinks/controller.coffee
@@ -100,7 +100,7 @@ class Turbolinks.Controller
   # Scrolling
 
   scrollToAnchor: (anchor) ->
-    if element = document.getElementById(anchor)
+    if element = document.getElementById(anchor) || document.querySelector("[name=#{anchor}]")
       @scrollToElement(element)
     else
       @scrollToPosition(x: 0, y: 0)
--- a/src/turbolinks/controller.coffee
+++ b/src/turbolinks/controller.coffee
@@ -100,7 +100,7 @@ class Turbolinks.Controller
   # Scrolling
 
   scrollToAnchor: (anchor) ->
-    if element = document.getElementById(anchor) || document.querySelector("[name=#{anchor}]")
+    if element = document.getElementById(anchor) || document.querySelector("a[name=#{anchor}]")
       @scrollToElement(element)
     else
       @scrollToPosition(x: 0, y: 0)


trostli:
I was actually looking for how to get this exact thing to work. This seems like a fantastic PR!

nateberkopec:
👍 because the name attribute is deprecated in HTML5, not removed (it generates validation warnings, not errors). 

I do think it might be useful to add a single test here.

It may be useful to extract into a named function at this point to improve readability:

```coffee
anchorIsScrollable (anchor) ->
    document.getElementById(anchor) || document.querySelector("a[name=#{anchor}]")
```

It may be faster to do this with a single querySelector call rather than doing an `||` here. jsperf is down so I can't check.

```coffee
anchorIsScrollable (anchor) ->
    document.querySelector("##{anchor}", "a[name=#{anchor}]")
```

trostli:
anything stopping this from being merged?

domchristie:
@trostli tests are needed, which I will add as soon as I can.

domchristie:
> It may be faster to do this with a single querySelector call rather than doing an || here

It might be faster to query named elements with a single querySelector, but for the common case with ID'd elements, the `getElementById(…) || querySelector(…)` seems [faster](https://jsperf.com/document-getelementbyid-vs-document-queryselector).

I am still having a little difficulty trying to test this behaviour, since it doesn't seem possible to check that the window has scrolled after navigation. Any ideas?

packagethief:
I noticed the same thing when testing, @domchristie. I couldn't figure it out.

I agree that using a single selector would be better. Here's how I approached it, plucked from an old stash:

```diff
diff --git a/src/turbolinks/controller.coffee b/src/turbolinks/controller.coffee
index cb89d8e..a6b35ef 100644
--- a/src/turbolinks/controller.coffee
+++ b/src/turbolinks/controller.coffee
@@ -100,7 +100,7 @@ class Turbolinks.Controller
   # Scrolling
 
   scrollToAnchor: (anchor) ->
-    if element = document.getElementById(anchor)
+    if element = @view.getElementForAnchor(anchor)
       @scrollToElement(element)
     else
       @scrollToPosition(x: 0, y: 0)
diff --git a/src/turbolinks/snapshot.coffee b/src/turbolinks/snapshot.coffee
index 5291c11..b975b18 100644
--- a/src/turbolinks/snapshot.coffee
+++ b/src/turbolinks/snapshot.coffee
@@ -31,8 +31,11 @@ class Turbolinks.Snapshot
   getCacheControlValue: ->
     @getSetting("cache-control")
 
+  getElementForAnchor: (anchor) ->
+    try @body.querySelector("[id='#{anchor}'], a[name='#{anchor}']")
+
   hasAnchor: (anchor) ->
-    try @body.querySelector("[id='#{anchor}']")?
+    @getElementForAnchor(anchor)?
 
   isPreviewable: ->
     @getCacheControlValue() isnt "no-preview"
diff --git a/src/turbolinks/view.coffee b/src/turbolinks/view.coffee
index 2fa2eb2..88f5c79 100644
--- a/src/turbolinks/view.coffee
+++ b/src/turbolinks/view.coffee
@@ -8,7 +8,10 @@ class Turbolinks.View
 
   getRootLocation: ->
     @getSnapshot().getRootLocation()
-    
+
+  getElementForAnchor: (anchor) ->
+    @getSnapshot().getElementForAnchor(anchor)
+
   getSnapshot: ->
     Turbolinks.Snapshot.fromElement(@element)
```

This is slightly different from your PR. `Turbolinks.Visit` also looks for an element matching the location's anchor [when deciding whether to use a cached version of the page](https://github.com/turbolinks/turbolinks/blob/f5f6495f650ab53d49e0e60554021b0472b299ec/src/turbolinks/visit.coffee#L50). This check should support named anchors too. The change above ensures we have a single way to get the element for an anchor, for both scrolling and caching.

If you agree, I could turn the stash above into a proper PR :smile:

domchristie:
> I noticed the same thing when testing, @domchristie. I couldn't figure it out.

I had a brief look into https://github.com/sstephenson/replicant, but like you, couldn't quite get it. I will give it another go at some point.

> If you agree, I could turn the stash above into a proper PR

Yes, this looks great! I meant to check how a restoration visit worked with an anchor, so this is nice 👍. 

Also, I wonder if this might be an opportunity to fix https://github.com/turbolinks/turbolinks/issues/75. I'm thinking we'll need to prevent the request but store a snapshot if that's possible(?). What do you think? I'd be happy to look into this if that'd help?

domchristie:
After a bit of digging, I have figured out a way to assert whether an element has been scrolled to. I will hopefully update #270 and #285 at some point this week.

domchristie:
Closed in favour of #348

PR #206, opened 2016-11-08
Add an action: noscroll to Turbolinks.visit

pedantic-git:
I had a use case in my app where part of the page reloads with Turbolinks and part of it doesn't (using data-turbolinks-permanent). In this situation, having the UI scroll back to the top on Turbolinks.visit was undesirable.

This PR adds a 'noscroll' action to Turbolinks.visit which can be called like this:

```javascript
Turbolinks.visit(location, {action: "noscroll"});
```

When called in this manner, a full Turbolinks visit takes place, but the window scroll position remains the same.


kirillplatonov:
👍 for this PR. Similar feature was in turbolinks-classic.

kirillplatonov:
@sstephenson @packagethief could you consider to merge this PR, please?

josepjaume:
I'm also in need of something like this, but I would rather see a `data-turbolinks-scroll="no"` attribute instead - mixing it with the actions doesn't feel like the way to go.

pedantic-git:
@josepjaume Unfortunately that doesn't solve my use case, which is in code that calls `Turbolinks.visit` directly. It's not a bad idea to add this as an additional way of accessing the code, though.

amnesia7:
This would be handy. 

In my case I would need it to be an option on `.visit()` since for me noscroll would depend on the function performed on the page rather than an attribute on a link or a meta tag. My page is interactive and scrolling to the top isn't always required; it depends on the interaction performed by the user.

amnesia7:
If not an `action` then there's always the way classic did it (`scroll: {boolean}`) - as an option alongside `action` rather than as an actual `action` variant.

pedantic-git:
Can I get some clarification on why people are concerned about it being mixed with `action`? All the current values of `action` are related to scrolling.

amnesia7:
@pedantic-git 
If it is another variant of `action` does that not prevent the action of `advance` or `replace` from being specified?
The `action` looks like it relates to the type of visit, not how the page should react when the user visits it.

pedantic-git:
Oh I see, because `advance` and `replace` also affect the history. Understood! I'll try to find time to adjust this.

rogierverbrugge:
I am in need of this functionality so started working on another way to implement this. Check out 

https://github.com/rogierverbrugge/turbolinks/commit/aa5a6d43b1419872378cd96c06ad1d85b9c6d842

This way you can use Turbolinks.visit(url, {scroll: 'noscroll'});

pedantic-git:
@rogierverbrugge This is amazing! I took a look for a way to do exactly this and then didn't have time. Are you happy to make another PR for this change and then I can close this one?

pedantic-git:
Closing this PR because #258 is better!

PR #217, opened 2016-12-11
Readme Table of Contents

brock-noah:
The readme has a lot of content but no high level navigation. Adds a table of contents with links to each header.


packagethief:
This is great, @brock-noah. Much easier to navigate. Thank-you!

PR #236, opened 2017-02-16
Fix link with download attr is visited

ypresto:
Clicking on `<a download="...">` will download linked file on modern browsers.
But currently turbolinks captures and visits linked url, so for example mp4 video will be played in current tab.
This PR fixes it by exclude a tag with download attr.


packagethief:
Thanks @ypresto!

packagethief:
For posterity, I added a test for `a[download]` links in e7c5392.

ypresto:
Looks good! Thanks...!

PR #258, opened 2017-04-12
Implemented a noscroll option

rogierverbrugge:
In reply to #206 i've created a different approach to handle the onscroll handler option.

This way you can use `Turbolinks.visit(url, {scroll: 'noscroll'});` for noscroll.

For now the back button makes the content to scroll to top but maybe someone else can point me in the right direction.

Works great for page-filters or calendar navigation.


panoply:
+1 on this merge.

inopinatus:
Thanks for doing this! turbolinks-classic has equivalent function via enableTransitionCache, so the absence was a regression.

panoply:
Is this available via `data-attribute` ?

PR #270, opened 2017-04-26
Improve handling of URLs with extensions and fragment identifiers

domchristie:
Turbolinks is currently ignoring URLs which include an extension as well as a fragment identifier. This is because it will only handle URLs that point to HTML resources, and in this case, the URL is wrongly interpreted as being non-HTML.

A `Location`'s [`isHTML`](https://github.com/turbolinks/turbolinks/blob/c23dbe1beb0f220ada7f422abbf08943a864066a/src/turbolinks/location.coffee#L36) method returns true if its URL ends with an HTML extension (htm, html, or xhtml), or if it has no extension at all. However, if the URL has an HTML extension _and_ a fragment identifier, `isHTML` will be false because it deems everything after the last `.` to be part of the extension. For example the extension of `http://example.com/index.html#foo` would be `.html#foo`. This does not match any of the acceptable extensions, and therefore `isHTML` will return false and the URL will not be handled.

I have added a demo of this behaviour here: https://thawing-hollows-51529.herokuapp.com/ Notice that clicking on the link to `page_2.html#foo` performs a full page load rather than a Turbolinks load.

The proposed solution in this pull request is to use the `requestURL`as the basis for retrieving the extension as it excludes the fragment identifier.

### DIFF
--- a/test/src/modules/navigation_tests.coffee
+++ b/test/src/modules/navigation_tests.coffee
@@ -39,8 +39,8 @@ navigationTest "following a same-origin data-turbolinks=true link inside a data-
     assert.equal(navigation.action, "push")
     done()
 
-navigationTest "following a same-origin link with ID hash", (assert, session, done) ->
-  session.clickSelector "#same-origin-id-link", (navigation) ->
+navigationTest "following a same-origin anchored link", (assert, session, done) ->
+  session.clickSelector "#same-origin-anchored-link", (navigation) ->
     session.waitForEvent "turbolinks:load", ->
       assert.equal(navigation.location.pathname, "/fixtures/one.html")
       assert.equal(navigation.location.hash, "#element-id")
--- a/test/src/modules/navigation_tests.coffee
+++ b/test/src/modules/navigation_tests.coffee
@@ -39,6 +39,14 @@ navigationTest "following a same-origin data-turbolinks=true link inside a data-
     assert.equal(navigation.action, "push")
     done()
 
+navigationTest "following a same-origin link with ID hash", (assert, session, done) ->
+  session.clickSelector "#same-origin-id-link", (navigation) ->
+    session.waitForEvent "turbolinks:load", ->
+      assert.equal(navigation.location.pathname, "/fixtures/one.html")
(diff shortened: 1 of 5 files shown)


domchristie:
Pull request description improved (above) with more detail about the issue.

test/src/modules/navigation_tests.coffee
```diff
@@ -39,6 +39,14 @@ navigationTest "following a same-origin data-turbolinks=true link inside a data-
     assert.equal(navigation.action, "push")
     done()
 
+navigationTest "following a same-origin link with ID hash", (assert, session, done) ->
```
packagethief: Turbolinks uses the term "anchor" internally. Let's stick with that convention and call this "following a same-origin anchored link" and `same-origin-anchored-link`.

domchristie: Cool. Updated in https://github.com/turbolinks/turbolinks/pull/270/commits/2aafe10f754c5c81d071ad6fdea5be2f81376984 (I think I went round in circles: fragment identifier, anchor, hash, fragment identifier, anchor, hash … :)

domchristie:
> Thank-you, especially for adding a test! Looks good to me.

Thanks! I actually came across this issue when attempting to test https://github.com/turbolinks/turbolinks/pull/175

domchristie:
Is there anything I should to add to this? Thanks.

javan:
Looks good to me. I think we should be using more of the properties from the `<a>` element instead of these regular expressions, but that's a separate issue.

screen shot 2017-06-09 at 1 27 29 pm

domchristie:
Hey @javan, do you have any thoughts on getting this (and maybe https://github.com/turbolinks/turbolinks/pull/285) merged? Thanks

javan:
@domchristie, I just requested a little review help. Feature-wise I'm +1 on getting both merged.

domchristie:
@sstephenson Thanks and no worries! It looks like [you have been busy](https://github.com/stimulusjs/stimulus) (amongst other things ;)

Issue #266, opened 2017-04-24
Turbolinks, async and DOMContentLoaded


nateberkopec:
See #28, where `defer` fixed the user issue but not really corrected Turbolinks behavior.

As pointed out by @domchristie in https://github.com/rails/rails/pull/23012#issuecomment-171804917, we can't make the default rails template application.js `async` or `defer` because [Turbolinks installs a handler of DOMContentLoaded](https://github.com/turbolinks/turbolinks/blob/af205f541c76fb01f6de7e1855362b888ba3c4e5/src/turbolinks/controller.coffee#L20), which may fire *after* Turbolinks execution if these attributes are used.

`async` and `defer` are really important attributes for front-end performance, and they're actually extremely compatible with how Turbolinks apps work. Since Turbolinks apps render HTML on the server, no Javascript is required to actually display the initial content, so it doesn't make sense that a simple Turbolinks app's JS bundle should block the browser HTML parser. Turbolinks should be designed to work well with `async` and/or `defer`.

A possible solution looks like checking [`document.readyState`](https://stackoverflow.com/questions/9237044/async-loaded-scripts-with-domcontentloaded-or-load-event-handlers-not-being-call) and I can open a PR to that effect if the maintainers agree that this looks like a solution.

nateberkopec:
Hit enter too quickly on this, text above now reflects what I actually wanted to say :)

Fudoshiki:
Any update?

javan:
Sounds reasonable to me. I think a change like this would fix it:
```diff
--- a/src/turbolinks/controller.coffee
+++ b/src/turbolinks/controller.coffee
@@ -17,7 +17,10 @@ class Turbolinks.Controller
   start: ->
     if Turbolinks.supported and not @started
       addEventListener("click", @clickCaptured, true)
-      addEventListener("DOMContentLoaded", @pageLoaded, false)
+      if document.readyState is "loading"
+        addEventListener("DOMContentLoaded", @pageLoaded, false)
+      else
+        @pageLoaded()
```

nateberkopec:
I'll throw a PR together.

packagethief:
Gah,

packagethief:
Oops, clicked close by accident!

Fudoshiki:
Hello, any progress?

kaspth:
@Fudoshiki feel free to volunteer the necessary PR if you're in a hurry 😉

Fudoshiki:
[...]

PR #274, opened 2017-04-30
If page is interactive, fire pageLoaded() immediately

nateberkopec:
When using Turbolinks with the `async` script attribute, Turbolinks may
execute after DOMContentLoaded. To fix this, we only attach a handler
to DOMContentLoaded if the readystate is not "loading" (that is to say,
it is "interactive" or "complete"). Otherwise, we fire the pageLoaded
event immediately without installing the handler.

This changes Turbolinks interaction with `defer` scripts. After this
change, an `async` Turbolinks script may execute *before* `defer`ed
scripts. This is because the browser executes all `defer` scripts *before*
the DOMContentLoaded event. Previously, Turbolinks would always execute
after these scripts, but may now execute before them in some cases
depending on when Turbolinks has finished downloading.

### DIFF
--- a/src/turbolinks/controller.coffee
+++ b/src/turbolinks/controller.coffee
@@ -17,12 +17,12 @@ class Turbolinks.Controller
   start: ->
     unless @started
       addEventListener("click", @clickCaptured, true)
+      @startHistory()
       if document.readyState is "loading"
         addEventListener("DOMContentLoaded", @pageLoaded, false)
       else
         @pageLoaded()
       @scrollManager.start()
-      @startHistory()
       @started = true
       @enabled = true
 
--- a/src/turbolinks/controller.coffee
+++ b/src/turbolinks/controller.coffee
@@ -17,7 +17,10 @@ class Turbolinks.Controller
   start: ->
     unless @started
       addEventListener("click", @clickCaptured, true)
-      addEventListener("DOMContentLoaded", @pageLoaded, false)
+      if document.readyState is "loading"
+        addEventListener("DOMContentLoaded", @pageLoaded, false)
+      else
+        @pageLoaded()
       @scrollManager.start()
       @startHistory()
       @started = true
--- a/README.md
+++ b/README.md
@@ -417,7 +417,7 @@ Turbolinks emits events that allow you to track the navigation lifecycle and res
 
 * `turbolinks:render` fires after Turbolinks renders the page. This event fires twice during an application visit to a cached location: once after rendering the cached version, and again after rendering the fresh version.
 
-* `turbolinks:load` fires once after the initial page load, and again after every Turbolinks visit. Access visit timing metrics with the `event.data.timing` object.
+* `turbolinks:load` fires once the first time Turbolinks is loaded, and again after every Turbolinks visit. Access visit timing metrics with the `event.data.timing` object.
 
 # Contributing to Turbolinks
 


nateberkopec:
Not sure the whole caveat about "now Turbolinks can execute before deferred scripts" even matters, but figured I'd mention it.

Closes #266

javan:
I've been thinking more on this and now I'm not sure what the right thing to do is. `turbolinks:load` more or less maps to when `DOMContentLoaded` would have fired in a non-Turbolinks app. With this change, `turbolinks:load` could fire any time initially depending on how and when Turbolinks is loaded. It could *also* fire if Turbolinks is stopped and then started again (e.g. `Turbolinks.controller.stop(); doSomething(); Turbolinks.controller.start()`)

nateberkopec:
> It could also fire if Turbolinks is stopped and then started again (e.g. Turbolinks.controller.stop(); doSomething(); Turbolinks.controller.start())

This could be easily fixed with just tracking the state somewhere to make sure it only goes off once, right?

> turbolinks:load could fire any time initially depending on how and when Turbolinks is loaded

This is interesting. Is this really a bad thing, though? If someone's just using JQuery and $(document).ready hooks, adding `async` would do the same thing.

javan:
> If someone's just using JQuery and `$(document).ready` hooks, adding `async` would do the same thing.

The difference is `$(document).ready` is a *function* that takes a callback, not an event. Events should indicate that something happened, and firing `turbolinks:load` at an indeterminate time *after* `DOMContentLoaded` breaks that contract.

nateberkopec:
It depends on what you think `turbolinks:load` means, I guess. If it means "the DOM is ready and Turbolinks has been evaluated" then nothing has changed. The only thing that's changed is "deferred scripts may not have been executed". 

> could fire any time initially 

To be clear, it can fire any time *after* the document has finished parsing (readystate `interactive`). It can't fire before that (still). It can also now fire before other deferred scripts. If I marked Turbolinks.js as `async`, "this event may fire as soon as it can, which may be after page load" is exactly the behavior I would expect.

sstephenson:
+1. Agree the initial `turbolinks:load` event should be dispatched immediately if the page is interactive.

I _don’t_ think apps should use `turbolinks:load` as a general-purpose hook for installing JS behavior, but I’ll accept this concession to the way most apps are likely using the event.

javan:
From the README:
> `turbolinks:load` fires once after the initial page load, and again after every Turbolinks visit.

This change assumes that turbolinks will always be loaded in the initial page's js (synchronously, `defer`'d, or `async`), but it's *possible* that it could be loaded in other ways. Imagine this pathological scenario:
```js
// <script src="main.js">
setTimeout(function() {
  lazyLoad("turbolinks.js")
}, 10000) 
```
Would you expect a `turbolinks:load` then?

Here's one way Turbolinks could handle it:
```diff
--- a/src/turbolinks/controller.coffee
+++ b/src/turbolinks/controller.coffee
@@ -17,9 +17,18 @@ class Turbolinks.Controller
   start: ->
     if Turbolinks.supported and not @started
       addEventListener("click", @clickCaptured, true)
-      addEventListener("DOMContentLoaded", @pageLoaded, false)
+      @startHistory() # Start history to set initial `@location`
+      switch document.readyState
+        when "loading"
+          # Page is still loading, fire `turbolinks:load` when it's interactive
+          addEventListener("DOMContentLoaded", @pageLoaded, false)
+        when "interactive"
+          # Page is interactive, fire `turbolinks:load` now
+          @pageLoaded()
+        else
+          # Page is loaded, don't fire `turbolinks:load`
+          @lastRenderedLocation = @location
       @scrollManager.start()
-      @startHistory()
       @started = true
       @enabled = true
```

nateberkopec:
I guess you're counting on the fact that the `turbolinks` javascript tag itself will keep the document in `interactive` state until it's finished executing? Is that really the case? [I don't think a Javascript subresource "stops delaying the load event"](https://dev.w3.org/html5/spec-preview/the-end.html)  while it's executing.

> Fetching an external script must delay the load event of the element's document until the task that is queued by the networking task source once the resource has been fetched (defined above) has been run.

So, a Turbolinks script with `async` and no other subresources on the page would probably hit the `else` condition here, which is probably not what you intended.

nateberkopec:
> Would you expect a turbolinks:load then?

My answer to that is "yeah, probably". I mean, what are the odds someone is inserting Turbolinks onto a page *and* counting on turbolinks:load not firing?

javan:
Cool, yeah, I'm just being difficult. 😬

src/turbolinks/controller.coffee
```diff
@@ -17,7 +17,10 @@ class Turbolinks.Controller
   start: ->
     unless @started
       addEventListener("click", @clickCaptured, true)
-      addEventListener("DOMContentLoaded", @pageLoaded, false)
+      if document.readyState is "loading"
+        addEventListener("DOMContentLoaded", @pageLoaded, false)
+      else
+        @pageLoaded()
```
javan: I think `@startHistory()` needs to be called before this condition so [`@location` is set](https://github.com/turbolinks/turbolinks/blob/master/src/turbolinks/controller.coffee#L59) for [`@pageLoaded()` to use](https://github.com/turbolinks/turbolinks/blob/master/src/turbolinks/controller.coffee#L138).

nateberkopec:
👍 moved startHistory

nateberkopec:
Proposing a README change as well to be more specific about when turbolinks:load fires, happy to workshop the phrasing.

javan:
👍 to the README change, but it's conflicting with master somehow. Mind fixing it up?

nateberkopec:
Removed the whitespace removal and it looks good now.

javan:
❤️

javan:
Turbolinks 5.0.1 is out!

glebm:
> My answer to that is "yeah, probably". I mean, what are the odds someone is inserting Turbolinks onto a page and counting on turbolinks:load not firing?

@nateberkopec The odds are 100% because that had to be accounted for to make Turbolinks 5.0.0 work with async scripts.

nateberkopec:
@glebm That comment was in regard to somebody injecting Turbolinks into a page long after initial load (i.e. not async or defer).

I still don't think we actually broke anything for you.

glebm:
> I still don't think we actually broke anything for you.

@nateberkopec What do you mean? It was working and now it's not (well I've patched it on master but tomorrow people will run `bundle update` and because you've only bumped PATCH it'll break for them).

maia:
@nateberkopec I'm also having issues with the new bahavior, and am unsure how to fix it. Until now I've loaded certain page elements that take longer to load due to db queries via ajax - I added `link_to`with `remote: true` and `class: "something"` to these elements, and on `turbolinks:load` I called `$('.something').click()`.

Now as `== javascript_include_tag 'application', 'data-turbolinks-track': 'reload', async: production?`frequently will load jQuery, turbolinks etc. *after* `DOMContentLoaded`, `turbolinks:load` does not always fire on initial page load.

If anyone has a suggestion how to properly load page elements automatically via ujs once the page itself is loaded, I'd be thankful. Using `DOMContentLoaded` to load them did cause side effects, as I also rely on other javascripts being loaded first. 

So the question is: which event can we rely on that will trigger *after* all javascript has been loaded async, even on the initial page load? Until turbolinks 5.0.0 it was `turbolinks:load`, what is it now?

nateberkopec:
@maia Upgrade to >= 5.0.2

mindreframer:
@nateberkopec latest published version as of today is  5.0.1 
https://rubygems.org/gems/turbolinks
screen shot 2017-07-28 at 16 01 49

??? is this still a not fixed bug?

nateberkopec:
@mindreframer https://rubygems.org/gems/turbolinks-source

mindreframer:
@nateberkopec great, thx!

PR #276, opened 2017-05-03
Fix DOM Ready check in IE10

nateberkopec:
I think this may warrant another patch release. 

It's really unclear to me exactly *how much earlier* IE10 can change to `interactive`. I'm guessing users with no `async` or `defer` attributes are unaffected.


nateberkopec:
Closing in favor of a more complete solution.

PR #284, opened 2017-05-04
Move call to .bind behind a Turbolinks.supported check

trueheart78:
Internet Explorer 8 continued to raise an exception when attempting to use Turbolinks v5.0.2. We kept getting, `TypeError: Object doesn't support this property or method` for `.bind()`, which is used only in the ScrollManager constructor. In this same browser, `Turbolinks.supported` returns `false`, so this was not expected.

This PR changes it to check that Turbolinks is supported prior to calling `.bind()`.

Worked with @jaketobin on this to verify the cause and that this change provides a fix.


nateberkopec:
Maybe this should short-circuit earlier, though? TL obviously isn't expected to work with IE8, so maybe it should shut off sooner.

packagethief:
I agree, @nateberkopec. `ScrollManager.start()` is already behind a [`Turbolinks.supported` check in the controller](https://github.com/turbolinks/turbolinks/blob/16df85c/src/turbolinks/controller.coffee#L18-L21). Can we wait until `start()` to create the throttled scroll handler so it needn't be done in the constructor?

trueheart78:
@nateberkopec The one other spot it could work is in the Controller class, where the ScrollManager is instantiated. There is a support check in the `start` method, but not in the constructor.

@packagethief Would that cause an issue with calling start/stop, since the instantiation would be in the start method?

packagethief:
@trueheart78, yes, I was suggesting we put it in ScrollManager's `start()`. Here's one way to do it:

```diff
diff --git a/src/turbolinks/scroll_manager.coffee b/src/turbolinks/scroll_manager.coffee
index 1de743e..5368ee6 100644
--- a/src/turbolinks/scroll_manager.coffee
+++ b/src/turbolinks/scroll_manager.coffee
@@ -1,16 +1,16 @@
 class Turbolinks.ScrollManager
   constructor: (@delegate) ->
-    @onScroll = Turbolinks.throttle(@onScroll).bind(this)
 
   start: ->
     unless @started
-      addEventListener("scroll", @onScroll, false)
+      @throttledOnScroll ?= Turbolinks.throttle(@onScroll).bind(this)
+      addEventListener("scroll", @throttledOnScroll, false)
       @onScroll()
       @started = true
 
   stop: ->
     if @started
-      removeEventListener("scroll", @onScroll, false)
+      removeEventListener("scroll", @throttledOnScroll, false)
       @started = false
 ```

I bet @javan can think of a better way 😄

**Comment by javan** [comment]
How about
```diff
diff --git a/src/turbolinks/scroll_manager.coffee b/src/turbolinks/scroll_manager.coffee
index 1de743e..98e38d1 100644
--- a/src/turbolinks/scroll_manager.coffee
+++ b/src/turbolinks/scroll_manager.coffee
@@ -1,6 +1,6 @@
 class Turbolinks.ScrollManager
   constructor: (@delegate) ->
-    @onScroll = Turbolinks.throttle(@onScroll).bind(this)
+    @onScroll = Turbolinks.throttle(@onScroll)
 
   start: ->
     unless @started
@@ -19,7 +19,7 @@ class Turbolinks.ScrollManager
   scrollToPosition: ({x, y}) ->
     window.scrollTo(x, y)
 
-  onScroll: (event) ->
+  onScroll: (event) =>
```

trueheart78:
Those both look like they would work. What should my next step be? Seeing commit hashes tells me you've both done some work.

packagethief:
@trueheart78, I'd update this PR using @javan's suggestion (and confirm it works). Thank-you!

trueheart78:
I updated it with @javan's suggestion, and force pushed. All tests still pass.

trueheart78:
IE8 is still working as-expected with the updated build.

packagethief:
Thanks @trueheart78!

trueheart78:
excited

Glad to help out, @packagethief @javan

packagethief:
Hey @trueheart78, not a big deal, but for future reference you don't need to commit the compiled JavaScript in `dist/turbolinks.js` when submitting a PR. We'll do that [when we cut a release](https://github.com/turbolinks/turbolinks/commit/16df85ca26aa4b77f02da48f24a84e36c79eb537). My bad for not noticing! 😄

trueheart78:
Thanks for the info, @packagethief.

And thanks [for the release](https://github.com/turbolinks/turbolinks/releases/tag/v5.0.3).

Issue #214, opened 2016-12-02
Turbolinks visit does not scroll to in-page anchors


bhox:
I have a bunch of anchors with inline links and when clicked turbolinks will reload the page, fetching the full url as a visit instead of scrolling to the anchor. 

For example, a link to `example.com/page#potato` will load `example.com/page#potato` instead of simply scrolling to the inline anchor.

I fixed it by adding `data-turbolinks="false" `to all inline anchor links, but I don't see any reason to not do this by default. Links with only hash fragments are inline and should be ignored by turbolinks (unless I'm missing something!).

benhoxie:
Similar problem when following a visit to a new page that includes an inline anchor via hash fragment. 

The page loads but does not scroll to a matching named element on page (the default html behavior).

benhoxie:
dupe of: https://github.com/turbolinks/turbolinks/issues/75

benhoxie:
Reopening for the second comment, following links with inline anchors. Only the first part is a dupe.

benhoxie:
I used this to scroll to inline anchors if they are present. It's not fancy.

`
$(document).on('turbolinks:load',function(event){
  // scroll to matching named element in hash
  if(window.location.hash){
    tags = document.getElementsByName(window.location.hash.slice(1))
    if(tags.length > 0){
      window.scroll(0,tags[0].offsetTop);
    }
  }

});
`

Tongboy:
Reading the cancelling visits section of the readme suggested a possible workaround - https://github.com/turbolinks/turbolinks#canceling-visits-before-they-start

In the `turbolinks:before-visit` event check if the url is the same as the current page and if so call `event.preventDefault()` to stop the visit and then scroll to the anchor.

thoughts?

joeljackson:
It would be nice to have a systematic recommended approach to this.

This has worked for me, but it feels like every use case of link I introduce I need to think through as opposed to it just working.

```
document.addEventListener('turbolinks:before-visit', function(e) {
  var parser = document.createElement('a');
  parser.href = e.data.url;

  if(parser.host == window.location.host && parser.pathname == window.location.pathname) {
    e.preventDefault();
    window.location.href = e.data.url;
  }
});
```

This idea is if the host and path have not changed maybe there is no need to trigger turbolinks, just use the browser behavior.

roooodcastro:
[...]

Issue #192, opened 2016-10-11
Turbolinks sends request to server when return back in browser


yuriti89:
Hi. I have following issue: when I click hashed anchor and after that return back in browser, turbolinks sends request to server. When I debug it, I detect that it happens because of entering "historyPoppedToLocationWithRestorationIdentifier" method.

It's not expecting behavior: I navigate back from the hashed url (http://example.com#hash) to the same url (http://example.com), why do I get request to server?

nateberkopec:
See #75.

Though your problem is a *slight* wrinkle on the original issue - when navigating from an anchor on a current page to the same page with *no* anchor, Turbolinks should ignore that navigation.

domchristie:
> I navigate back from the hashed url (http://example.com#hash) to the same url (http://example.com), why do I get request to server?

It seems like browsers will always issue requests to same-page anchors _unless_ the `href` starts with a "#". https://github.com/turbolinks/turbolinks/issues/285 implements the following:

- `http://example.com#hash` to `http://example.com` will issue a request
- `http://example.com#hash` to `http://example.com#` will _not_ issue a request

PR #285, opened 2017-05-04
Prevent issuing requests to same-page anchors

domchristie:
This is a stab at fixing #75, #214, and #192

## The Problem

Clicking on a link which references an anchor on the current page issues an unnecessary request. This can cause problems when linking to content which is lazily loaded (see https://agile-reef-88023.herokuapp.com/posts/1), as well as with [analytics](https://github.com/turbolinks/turbolinks/issues/75#issuecomment-255867311). [Disabling Turbolinks completely on those links](https://github.com/turbolinks/turbolinks/issues/75#issuecomment-244915109) prevents the request, but because the view is not cached, this approach breaks the Back button.

## The Solution

This PR detects whether the visit is to an anchor on the same page, and if so, prevents the request. It takes a snapshot and performs the scroll, bypassing any rendering and `turbolinks:load` triggering. A similar flow handles restorations to same-page anchors.

## Notes

This PR also tidies up `Location#anchor` so that a location with a blank anchor (e.g. `http://www.example.com#` will have an `anchor` property of `''`. This means that transitioning from `http://www.example.com#foo` to `http://www.example.com#` will not issue an unnecessary request (related: https://github.com/turbolinks/turbolinks/issues/192)

Same-page anchor links with `data-turbolinks=false` will still break the Back button since snapshotting is bypassed. I guess there is an argument that we make an exception in this case so that same-page anchors with `data-turbolinks=false` are handled as above to prevent this issue. The downside is that `data-turbolinks=false` is conditional and therefore may be confusing.


src/turbolinks/browser_adapter.coffee
```diff
@@ -14,6 +14,7 @@ class Turbolinks.BrowserAdapter
   visitStarted: (visit) ->
     visit.issueRequest()
     visit.changeHistory()
+    visit.goToSamePageAnchor()
     visit.loadCachedSnapshot()
```
domchristie: Calling `visit.goToSamePageAnchor` _and_ `visit.loadCachedSnapshot` on a restoration visit will result in `@adapter.visitRendered?(this)` being called twice. I think we may need a condition in either of those methods to prevent this.

domchristie:
Test added in https://github.com/turbolinks/turbolinks/pull/285/commits/f973aa127922a2390ddf4ae1676a963ed8ed8524 which should pass once https://github.com/sstephenson/replicant/issues/1 is resolved

domchristie:
I've attempted a fix for the replicant issue https://github.com/sstephenson/replicant/pull/2. The replicant build is available here: https://github.com/domchristie/replicant/blob/fix_same_page_anchor_location_changes_with_build/dist/replicant.js. Using it with this repo seems to fix the failing test :)

jondavidchristopher:
Anything I can do to help get this pull request merged @domchristie ?

domchristie:
Thanks for the nudge! I suppose it could too with a review from a maintainer. Otherwise if you fancy doing some manual testing of this branch, that'd be appreciated :)

src/turbolinks/browser_adapter.coffee
domchristie: I think I tested this and it doesn't seem to call `@adapter.visitRendered?(this)` twice. I will test again and confirm.

domchristie: Conveniently enough, Visit's `render` method cancels any existing render calls, and so `@adapter.visitRendered?(this)` will never be called twice (the `visit.loadCachedSnapshot` call will override that in `visit.goToSamePageAnchor`) 🎉

domchristie:
Having revisited the code, and performed some manual tests, I think this PR is in a reviewable state :) I'd welcome any feedback on the code design/approach. Thanks!

I will create an app demonstrating the fixes soon.

domchristie:
The following app demonstrates the fixes.

**[App with same-page anchor fix](https://agile-reef-88023-fix.herokuapp.com/posts/1)**

… and for comparison, [the same app with the buggy behaviour](https://agile-reef-88023.herokuapp.com/posts/1).

It is a simple blogging app. Comments are via AJAX after the post loads. When clicking on "Jump to Comment 1" in the buggy version, the page reloads, and attempts to jump to the anchor. However, at that point, the comments have not yet loaded, so the viewport does not scroll. This is fixed in the updated version.

src/turbolinks/location.coffee
```diff
@@ -6,17 +6,16 @@ class Turbolinks.Location
...
 
-    anchorLength = linkWithAnchor.hash.length
-    if anchorLength < 2
-      @requestURL = @absoluteURL
+    @requestURL = if @anchor?
+      @absoluteURL.slice(0, -(@anchor.length + 1))
     else
-      @requestURL = @absoluteURL.slice(0, -anchorLength)
-      @anchor = linkWithAnchor.hash.slice(1)
+      @absoluteURL
```
javan: I'm having a hard time digesting the changes here. Can you summarize them? Why switch from `link.hash` to a regex?

domchristie: This change is a workaround for the (inconsistent) native behaviour of `link.hash`. The native behaviour is as follows:

```js
link.href = '#foo'
link.hash // '#foo'

link.href = '#'
link.hash // ''

link.href = 'http://example.com'
link.hash // ''
```

This makes it impossible to distinguish whether a location has an empty hash or no hash at all.

To fix this, the `anchor` is just the text part of the hash or `null` if the URL does not contain a hash at all:

```js
link.href = '#foo'
@anchor // 'foo'

link.href = '#'
@anchor // ''

link.href = 'http://example.com'
@anchor // null
```

This is required for the [`locationHasSamePageAnchor`](https://github.com/turbolinks/turbolinks/pull/285/files#diff-511f2c4262a37015ef9a6d1a13d4a36dR163) method which calls `@location.anchor?`. Without this change, `@location.anchor?` will always be `true`.

In terms of the outcome, this change prevents Turbolinks from making requests when navigating to '#' (https://github.com/turbolinks/turbolinks/issues/192).

Let me know if you'd like me to add comments to the code for this.

domchristie: Hey @javan, do you have any thoughts on this approach?

javan: Cool, thanks for the explanation! 

Splitting the constructor code out into intention revealing methods/helpers would help clarify. Your changes don't make the code any more confusing so I'm not suggesting you should change anything. Just trying to grok the before / after.

domchristie: Yes, I’d happily tidy this up at some point; perhaps at the same time as refactoring the methods to use more `<a>` properties (as mentioned here: https://github.com/turbolinks/turbolinks/pull/270#issuecomment-307450179). Let me know if there’s anything else this PR needs. Thanks!

bragboy:
Any plans to merge this PR ?

domchristie:
@sstephenson I have rebased this against master and pushed. There is a failing test which I believe will be fixed by https://github.com/sstephenson/replicant/pull/2

javan:
Hey @domchristie, sorry for letting this sit for so long! It's a surprisingly thorny problem because we also need to support the native adapters: [turbolinks-ios](https://github.com/turbolinks/turbolinks-ios) and [turbolinks-android](https://github.com/turbolinks/turbolinks-android).

@sstephenson and I took a little time to recap:

In general, we want clicking anchors to behave exactly the way it would without Turbolinks involved. So each hash change adds a new history entry, and navigating back and forth renders and scrolls as necessary.

Scenarios that need to work:
1. On page 1, following a same-page anchor should scroll. Following a subsequent link to page 2 and then using the browser's back button should render page 1. Pressing the back button a second time should scroll.
2. We need to make sure the native adapters can handle anchor visits. We don't want existing apps to push a new view controller onto the stack each time you follow a same-page anchor.

I'm pretty sure your change addresses scenario 1, but not 2.

domchristie:
Thanks @javan and @sstephenson for reviewing this.

Yes, I think I have the first scenario covered. In terms of addressing the second scenario, how might you suggest approaching this?

Thanks again!

domchristie:
@javan @sstephenson I have given it a go on turbolinks-ios: https://github.com/turbolinks/turbolinks-ios/pull/126 let me know what you think :)

src/turbolinks/location.coffee
domchristie: I think we could tidy this up with:

```coffeescript
[@requestURL, @anchor] = @absoluteURL.split('#')
```

mrfoto:
Any plans to ever merge this PR?

chrickso:
please merge this

Fudoshiki:
What's status?

domchristie:
> What's status?

I have just rebased this so it should be up-to-date with the current master. If anyone needs a transpiled version, let me know.

Fudoshiki:
@domchristie yes, please

Fudoshiki:
@domchristie what with failing test?

domchristie:
> failing test?

Turbolinks uses [replicant](https://github.com/sstephenson/replicant) to simulate navigation in tests, however there was an issue when testing visits to same-page anchors (https://github.com/sstephenson/replicant/issues/1). This issue was fixed in https://github.com/sstephenson/replicant/pull/2 but I have just noticed that the repository has been archived, so I'm not sure where we go from here.

@sstephenson are you able to shed any light on this?

richardonrails:
What's needed in order to merge this?

domchristie:
> What's needed in order to merge this?

The tests have been updated to the new framework, but in the meantime, you may want to try [this workaround](https://github.com/turbolinks/turbolinks/issues/75#issuecomment-443256173)

euxx:
Hi @domchristie, thanks to your effort on the PR, and it would be great if we could managed to merge this PR after such a long time. : )

Issue #115, opened 2016-06-15
Problem with turbolinks:load including initial page load


pbougie:
I'm having trouble adding page tracking with [Gauges](http://get.gaug.es/) when using Turbolinks. Gauges automatically tracks a page when the script is embedded in a page. Therefore, placing it in the body tracks each page twice when a previously cached page is visited. Moving the script into the head section tracks the page on initial page load but not for Turbolinks visits.

If we add a little code, we can also track Turbolinks visits but there is a problem with the events. Using `turbolinks:load` will track the initial page load twice. Using any event before `turbolinks:load` will track the page twice in certain instances or track the wrong page title.

Maybe a new event that does not include the initial page load is required? Or a way to exclude the initial page load from `turbolinks:load`. Is there a verification in JS that can tell us if the page was loaded using Turbolinks? If so, I could use `turbolinks:load` and simply exclude the tracking on initial page load.

Rails: `4.2.6`
Turbolinks: `5.0.0.beta5`

packagethief:
Hey @pbougie. One way to detect the initial page load is to inspect the `event` data on `turbolinks:load`. On the initial load, `event.data.timing.visitStart` will be undefined.


```js
document.addEventListener("turbolinks:load", function(event) {
  if (!event.data.timing.visitStart) {
    // Initial page load
  } else {
    // Turbolinks page load
  }
})
```

Of course, this is totally not obvious and feels like a hack. It would be much nicer if we had an `event.data.initialPageLoad` boolean so we can be explicit.

Is that something you're interested in exploring in a PR?

pbougie:
Hi @packagethief. Looks like I got it working based on your suggestion. Here's what I did:

```js
document.addEventListener('turbolinks:load', function(event) {
  if (event.data.timing.visitStart) {
    _gauges.push(['track']);
  }
})
```
I only need the method to be executed when the page is loaded using Turbolinks. Not sure I understand when the visit timing metrics are started. Are `event.data.timing.visitStart` and `turbolinks:visit` related in any way in the lifecycle?

Not sure `event.data.initialPageLoad` would be less of a hack. Wouldn't an event that excludes the initial page load make more sense? I'm currently migrating an app from Pjax and I was a little confused at first that `turbolinks:load` included the `ready` event.

packagethief:
In most cases you don't need to distinguish between the initial page load and subsequent Turbolinks loads. That's why there's a single `turbolinks:load` event.

[...]

PR #286, opened 2017-05-06
Add flag to indicate initial page loads

benningm:
This adds a boolean flag event.data.initialPageLoad to the
turbolinks:load event to indicate if it is fired on
and initial page load.

### DIFF
--- a/src/turbolinks/controller.coffee
+++ b/src/turbolinks/controller.coffee
@@ -136,7 +136,7 @@ class Turbolinks.Controller
 
   pageLoaded: =>
     @lastRenderedLocation = @location
-    @notifyApplicationAfterPageLoad()
+    @notifyApplicationAfterPageLoad(undefined, true)
 
   clickCaptured: =>
     removeEventListener("click", @clickBubbled, false)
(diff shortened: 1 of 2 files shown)


nateberkopec:
👍 And it even has a test! 🎉

Fudoshiki:
Will fire only once on first turbolinks load?
Why need this PR?
For Turbolinks ready you can just check window.Turbolinks on existence or am I wrong and it's not about that?

nateberkopec:
@Fudoshiki Check #115. It adds a property to the event, it doesn't change when the event fires.

src/turbolinks/controller.coffee
```diff
@@ -179,8 +179,8 @@ class Turbolinks.Controller
   notifyApplicationAfterRender: ->
     Turbolinks.dispatch("turbolinks:render")
 
-  notifyApplicationAfterPageLoad: (timing = {}) ->
-    Turbolinks.dispatch("turbolinks:load", data: { url: @location.absoluteURL, timing })
+  notifyApplicationAfterPageLoad: (timing = {}, initial = false) ->
```
packagethief: Instead of positional parameters, let's use an object so we can name its keys. This lets us avoid the awkward `@notifyApplicationAfterPageLoad(undefined, true)` construct.

```coffee
notifyApplicationAfterPageLoad: ({timing, initialPageLoad} = {}) ->
  Turbolinks.dispatch("turbolinks:load", data: { url: @location.absoluteURL, timing, initialPageLoad })
```

Now we can write this:
```coffee
@notifyApplicationAfterPageLoad(initialPageLoad: true)
```

Fudoshiki:
@nateberkopec will it add property for each initial load for every new page or just one time on turbolinks script load?

nateberkopec:
[The property is always set](https://github.com/turbolinks/turbolinks/pull/286/files#diff-97de05b8058cb436109ef527e9cddc10R183) - true when it's initial page load, false on each Turbolinks-enabled navigation (anchor tag click).

benningm:
Just pushed the changes to use named parameters

davidsomen:
What's happening with this PR? I would really appreciate a feature like this. I need it to cause an animation to fire on the initial page visit.

abinoda:
Would love to see this 🚢

PR #304, opened 2017-07-13
fix: unsupported browser (IE8)

davegudge:
`Turbolinks.visit` states:

>If location is a cross-origin URL, or falls outside of the specified
root (see Setting a Root Location), or if the value of
Turbolinks.supported is false, Turbolinks performs a full page load by
setting window.location.

However this did not appear to be true for the IE8 (unsupported browser).

This commit allows `Turbolinks.visit` to gracefully fallback to the
standard javascript redirect via setting `window.location`.


davegudge:
Closing - IE8 is redundant now.

PR #316, opened 2017-08-24
Remove document.readyState as pageIsLoaded criteria

dennisreimann:
As [described here](https://github.com/turbolinks/turbolinks/commit/e90e4c4ad94b35d9f3f0a90b6810aa65311175d5##commitcomment-23820789) the `document.readyState is "complete"` breaks back button functionality in Safari >= 10.

Also moved and extended the comment to make it a bit clearer what is happening.

### DIFF
--- a/src/turbolinks/history.coffee
+++ b/src/turbolinks/history.coffee
@@ -23,6 +23,14 @@ class Turbolinks.History
 
   # Event handlers
 
+  # Chrome < 34 and Safari < 10 dispatch an initial popstate event on page load.
+  # Ignore it by setting the pageLoaded after the page load is done (hence defer).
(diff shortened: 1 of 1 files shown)


javan:
The `document.readyState is "complete"` is just a guard for the case where Turbolinks may have *after* the `load` event. Without it we'd never handle `popstate`.

It seems that Safari has a number of quirks with `popstate` and fixing this issue will require a different approach.

References:
- https://github.com/devote/HTML5-History-API/issues/69
- https://github.com/yahoo/fluxible/issues/349
- https://github.com/visionmedia/page.js/issues/213
- https://bugs.webkit.org/show_bug.cgi?id=93506

PR #325, opened 2017-09-22
Added "after-cache" event.

### LINKED ISSUE #324: Need an "after-cache" event.
I have a unique scenario for my app where I need an event that fires after the cache snapshot is taken, but not necessarily after the XHR request is complete.

AndBobsYourUncle:
Fixes #324 

Proposed flow:

1. User navigates to different page.
2. Snapshot is taken.
3. Some code runs that manipulates the DOM, because you listen to the new "after-cache" event.
4. The XHR request completes, and the normal flow resumes as usual.

In this way, any DOM manipulation in code that is listening for the "after-cache" event is not cached, but it is also immediately viewable before the XHR request is done.

This enables quite a few things that might otherwise be quite cumbersome to implement. Namely, you could do some animations for the page transition. Or you could even have a nav bar that responds immediately, even before the XHR request is complete.


risinglf:
Hi guys, what do you think about working and merging this PR?

kat0d:
so it's not merged for v5.2.0 ?

PR #331, opened 2017-11-08
Scrolling to an anchor on a redirect doesn't work

balvig:
## Problem

Currently, if you do something like:

```ruby
def show
  redirect_to root_url(anchor: "home_tabs")
end
```

and visit the page with turbolinks, the browser will fail to scroll down to the specified anchor:

## Cause

The cause seems to be that the callback `performScroll()` gets run _before_ `followRedirect()` has a chance to update `@location` to the final URL:

```coffee
loadResponse: ->
  ...
  @controller.render(snapshot: @response, @performScroll) # adds @performScroll as callback
  ...
  @complete() # triggers @adapter.visitCompleted() which triggers followRedirect()

followRedirect: ->
  if @redirectedToLocation and not @followedRedirect
    @location = @redirectedToLocation # Location is updated, but too late
    ...

scrollToAnchor: ->
  if @location.anchor? # @location is still the URL _before_ the redirect here
    @controller.scrollToAnchor(@location.anchor)
```

## Proposed solution

I admit to not fully knowing the background of needing to trigger `visitRendered` and `visitCompleted` on the adapter, so the proposal may be way off point, but basically it ensures that `followRedirect()` is called _before_ any scrolling is done, by calling it directly from within `loadResponse` before `@controller.render()`.

(I'm also not sure how to go about adding a test for this, as it kind of requires a server side component, so forgive the test-less PR 🙇 )

### DIFF
--- a/src/turbolinks/browser_adapter.coffee
+++ b/src/turbolinks/browser_adapter.coffee
@@ -39,9 +39,6 @@ class Turbolinks.BrowserAdapter
   visitRequestFinished: (visit) ->
     @hideProgressBar()
 
-  visitCompleted: (visit) ->
-    visit.followRedirect()
-
   pageInvalidated: ->
     @reload()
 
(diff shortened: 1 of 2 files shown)


balvig:
Fixed by https://github.com/turbolinks/turbolinks/pull/348 looks like, thanks a lot! 👍

PR #337, opened 2017-12-15
Improve compatibility with webcomponents.js (connectedCallback)

denisw:
### Background

[Custom Elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Custom_Elements) are a great fit for Turbolinks (and are rightfully mentioned in the Turbolinks README). Unfortunately, more often than not it's still required to use the webcomponents.js [Custom Elements polyfill](https://github.com/webcomponents/custom-elements) to support browsers other than Chrome and Safari 11.

This, in itself, is not much of a problem. However, I noticed that on browsers that need to fall back to the polyfill, my custom elements' `connectedCallback` methods are not called after following a link, breaking them all. 😞 

### The Problem

When visiting a new page, Turbolinks currently replaces the body by setting `document.body`:

```coffeescript
  assignNewBody: ->
    document.body = @newBody
```

However, the polyfill cannot detect `document.body` changes (https://github.com/webcomponents/custom-elements/issues/126). So it will not be able to walk the DOM and call the `connectedCallback`methods of all custom elements.

### The Solution

We can make the webcomponents.js polyfill aware of the body change by simply using `Element.prototype.replaceChild` instead:

```coffeescript
  assignNewBody: ->
    document.documentElement.replaceChild(@newBody, document.body)
```

This version (implemented by this PR) is totally equivalent to setting `document.body`, but `replaceChild` is patched by the polyfill, so `connectedCallback` works as expected. 

There are no disadvantages to this change. As `replaceChild` is supported on all browsers including IE back to version 6, no new browser incompatibilities are introduced.

### DIFF
--- a/src/turbolinks/snapshot_renderer.coffee
+++ b/src/turbolinks/snapshot_renderer.coffee
@@ -58,7 +58,7 @@ class Turbolinks.SnapshotRenderer extends Turbolinks.Renderer
       replaceableElement.parentNode.replaceChild(element, replaceableElement)
 
   assignNewBody: ->
-    document.body = @newBody
+    document.documentElement.replaceChild(@newBody, document.body)
 
   focusFirstAutofocusableElement: ->
     @findFirstAutofocusableElement()?.focus()


javan:
Thanks for the heads up! Your patch to fix the the polyfill seems like the best overall solution: https://github.com/webcomponents/custom-elements/pull/127.

seanpdoyle:
According to https://github.com/webcomponents/custom-elements/pull/127#issuecomment-357092320, @denisw 's change was merged, reverted, then re-opened as https://github.com/webcomponents/custom-elements/pull/134.

Unfortunately, fixing this WebComponents-side seems to be blocked.

@javan is this proposed change worth reconsidering?

sstephenson:
We'll need to test it carefully, of course, but I don't see any issue with using `replaceChild`.

javan:
+1, there's a new `replaceElementWithElement()` helper that was added after this PR was opened: https://github.com/turbolinks/turbolinks/blob/c340c42e2ece8c057aec611074971462a6e76c71/src/turbolinks/snapshot_renderer.coffee#L102-L104
We should probably use it instead of the current implementation for consistency.

seanpdoyle:
@sstephenson @javan I've opened https://github.com/turbolinks/turbolinks/pull/431 so that the CI test suite can run on then change, and on the off-change that this PR stays inactive.

If it'd be better for us to centralize the work to this PR, I'll gladly close https://github.com/turbolinks/turbolinks/pull/431.

sstephenson:
Thank you! https://github.com/turbolinks/turbolinks/releases/tag/v5.3.0-beta.1

PR #340, opened 2017-12-27
Fix IE 10 and IE 11's broken `preventDefault` [Fixes #233]

### LINKED ISSUE #233: Cancelling a Turbolinks visit before is starts fails in IE10/11
## Issue

Cancelling a Turbolinks visit before is starts fails in IE10/11 because - I think - IE10/11 .preventDefault doesn't work with "synthetic" events properly.

I have assumed that IE11 at least is still worth supporting at this point as Edge still has a small user base (mind you, I just looked and IE is running at about 5% these days! 😮 )

## Expected behaviour:

As described [in the readme](https://github.com/turbolinks/turbolinks#canceling-visits-before-they-start), if I add an event listener to handle an event of type `turbolinks:before-visit` then an `event.preventDefault()` should stop the visit from happening.

## Actual behaviour

I've create a [minimal example](https://rosswintle.uk/turbolinks-bug/).  If you open that and click the link in a 'modern' browser then the code intercepts the `before-visit` event, and `preventDefault()` cancels it.

However, if you open this link in IE10 or 11 (I've not tested any earlier versions) then the event is handled, and the `defaultPrevented` attribute of the event is set correctly, but Turbolinks fails to prevent the load.

## Analysis

I've spent a good few hours with this and from what I can see `preventDefault()` should, and indeed does, set the `defaultPrevented` attribute to true. However, on exit from the event handler, `defaultPrevented` is set back to false!

It's been hard to track down much info on this, but there's a few references where this has come up:

* A couple of StackOverflow threads, for example [this one](http://stackoverflow.com/questions/23349191/event-preventdefault-is-not-working-in-ie-11-for-custom-events)
* Bizarre workarounds in other people's code, [the `createEvent` method here, for example](https://academy.terrasoft.ua/jscoresdk/source/Event2.html)
* [This issue](https://github.com/webcomponents/webcomponentsjs/issues/157) in WebComponents (and [its fix](https://github.com/webcomponents/webcomponentsjs/commit/ff1ba94f92efa3b882cc51668d4cfec0acac0e92))

I've tried a few things to try and overcome this - like replicating some of the workarounds mentioned above. But I don't seem to be able to 'override' preventDefault the way other people have and I've been unsuccessful in all my attempts to get this to work.

Note: I'm not a JS Guru - I'd say my JS knowledge is 7/10. I'm not very familiar with coffeescript, but I think I've been able to make sense of it.

The relevant code seems to be:

* [`notifyApplicationBeforeVisitingLocation` in `controller.coffee`](https://github.com/turbolinks/turbolinks/blob/master/src/turbolinks/controller.coffee#L167) which 
dispatches the event
* [`applicationAllowsVisitingLocation` in `controller.coffee`](https://github.com/turbolinks/turbolinks/blob/master/src/turbolinks/controller.coffee#L160) which calls `notifyApplicationBeforeVisitingLocation` and returns `not event.defaultPrevented`

In my [minimal example](https://rosswintle.uk/turbolinks-bug/) you'll note that I've used a compiled-but-uncompressed copy of Turbolinks and added some logging.  In IE, my console logging outputs something like:

```
defaultPrevented: true
in applicationAllowsVisitingLocation. defaultPrevented is false
```

demonstrating that the `defaultPrevented` is changed back to false when exiting the `before-visit` event handler.

I don't know what we do about this. Is there a workaround that other people know of that I don't? Or do we need to create a more explicit mechanism for cancelling the event that works across browsers?  Perhaps setting `event.data.cancelVisit = true` and then detecting that instead of `event.defaultPrevented`?  I've provided [a non-coffeescript prototype of this that seems to work correctly](https://rosswintle.uk/turbolinks-bug/index.fixed.html) - the diff of this is:

```
1719c1719
<           return !event.defaultPrevented;
---
>           return !event.data.cancelVisit;
1735c1735,1736
<               url: location.absoluteURL
---
>               url: location.absoluteURL,
>               cancelVisit: false
```

I'm sure that wouldn't be hard to translate back into coffeescript. We'd need to make it backwards compatible I guess, using `defaultPrevented` if the new data attribute wasn't available, and we'd need to update the README too.

Phew. What do you think? Can we fix this?

Fudoshiki:
Fix #233


Fudoshiki:
Hello @nateberkopec, can review and merge this PR, already merged in Rails https://github.com/rails/rails/pull/31574

PR #343, opened 2018-01-04
Warn when loading Turbolinks from a <script> inside <body>

sstephenson:
This branch shows an early warning in the console when `document.currentScript` is a child of `<body>`:

screen shot 2018-01-04 at 8 33 43 am

Turbolinks is designed for application scripts to live in `<head>`, but best practices from several years ago encouraged folks to put all scripts at the end of `<body>`. That seems to trip people up fairly often.

### DIFF
--- a/src/turbolinks/script_warning.coffee
+++ b/src/turbolinks/script_warning.coffee
@@ -1,0 +1,16 @@
+do ->
+  return unless element = script = document.currentScript
+  return if script.hasAttribute("data-turbolinks-suppress-warning")
+
+  while element = element.parentNode
+    if element is document.body
+      return console.warn """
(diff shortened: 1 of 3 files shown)


src/turbolinks/script_warning.coffee
```diff
@@ -0,0 +1,16 @@
+do ->
```
javan: Each `.coffee` file has an implicit `do ->` so this adds an unnecessary second.

sstephenson: Technically unnecessary, but I wanted to make it clear the three `return` statements short-circuit just the script check and not the whole Turbolinks boot process.

PR #373, opened 2018-03-28
Update gem installation instructions

NoelDeMartin:
The new version of the ruby gem (5.1.0) allows using turbolinks without the asset pipeline, as discussed here: https://github.com/turbolinks/turbolinks-rails/pull/18

I think it's important to update this documentation, because installing an older version without the asset pipeline throws an error.


sstephenson:
Thank you!

Issue #43, opened 2016-03-10
Replace a dom element other than body?


wayneashleyberry:
Would it be possible to add an option to replace something other than the `<body>`?

I've built many site where the header / footer needed to stick around and only the content inbetween needs to be replaced. I've tried using `data-turbolinks-permanent` in beta 3 but it doesn't seem to have the same effect.

I'm thinking something like this:

```html
<head>
  ...
  <meta name="turbolinks-container" content="#main">
</head>
```

packagethief:
Hi Wayne,

Part of what makes Turbolinks so simple is that it "just" swaps the `<body>`. I don't think we should deviate from that.

I'm curious why permanent elements aren't working for you. Could you elaborate?

wayneashleyberry:
Thanks for responding @packagethief. I love the default zero-configuration behaviour of swapping the contents of the body, it's amazing!

I've been trying to use permanent elements to persist dom created by 3rd party javascript, which I have no control over unfortunately. That hasn't been working for me. I think it might be because turbolinks is looking for `data-turbolinks-permanent` once, on initialisation and I'm trying to dynamically add the attribute later on?

packagethief:
> I think it might be because turbolinks is looking for data-turbolinks-permanent once, on initialisation and I'm trying to dynamically add the attribute later on?

That sounds like the problem. Permanent elements must match between pages. If you add one with JavaScript that isn't present on the next page rendered from the server, it won't be carried over.

What about creating a permanent container element that you do render on the server, and placing your 3rd party DOM inside it?

wayneashleyberry:
That's the thing though @packagethief, I'm currently facing a scenario where I can't tell the javascript where to render it's dom :(

fernandes:
@packagethief I got a similar problem using the [toastr](https://github.com/CodeSeven/toastr) plugin

the idea was bump a toastr after the model was successfully created, so on `create.js.erb` I just triggered and everything should work fine.

My case was similar to @wayneashleyberry one, once I don't have control about the rendering, then my solution was

```javascript
$( document ).one( "page:load", function() {
  toastr.success("message", "title", {timeOut: '5000', closeButton: true, progressBar: true});
});
Turbolinks.visit('/path/to/models');
```

MSchmidt:
[...]

PR #401, opened 2018-07-09
Feature root element

### LINKED ISSUE #400: Turbolinks breaks ad viewability trackers
We use Turbolinks on our ad-supported site and have recently suffered from a bug that prevents ad viewability trackers to fire after Turbolinks visits. Specifically, it is the [Active View](https://support.google.com/dfp_premium/answer/6123557?hl=en) tracker that DFP uses.

Removing Turbolinks, we see ads behaving as expected. We also have some pages that refresh ads throughout the lifetime of the single page; these ad loads behave as expected, too. Only on Turbolinks visits do these trackers not fire properly.

This is a somewhat recent development. It seems correlated with a GPT.js release, and we're reaching out to Google to see if there is anything they are aware of on their end that may have broken Turbolinks usage.

Our hypothesis is that Turbolinks destroying the whole `<body>` tag between page views is the culprit. I have tried adding an opt-in feature to Turbolinks that only swaps a portion of the `<body>`. This seems to resolve the issue we are facing, though it's not entirely clear yet if this might be introducing other subtle bugs.

Has anybody else come across this specific issue with Active View trackers? Or similar issues with third party code that breaks when the `<body>` is wholesale replaced?

ahaurw01:
Fixes #400 

New opt-in feature: annotating a new page's element with `data-turbolinks-root` will swap in only that element and not the entire `body`, provided that the previous page also had a `data-turbolinks-root` element.

This is an attempt to play nice with DFP's Active View tracker which seems to break when the entire `body` is replaced between page views.

This feature is sort of the opposite side of the coin of `data-turbolinks-permanent`. There would seem to be plenty of cases where third party logic inserts elements in the DOM that you'd want to carry over to future page views.

Has discussion of this feature come up previously? Any reasons that this is obviously a bad idea?

This feature seems to break snapshot caching since the body is mutated instead of plucked away entirely. Any advice on a simple way to allow caching to occur with this new feature?


MidnightWonderer:
Should `data-turbolinks-root` be a selector configuration passed to `Turbolinks.start()` instead?

boqusi:
That would be very helpful. Turbolinks doesn't let my animations (e.g. closing a drawer) finish because it replaces the whole document.body, and using this fork is currently the only solution I got to work.

domchristie:
Hi @ahaurw01 thanks for this. Replacing a custom DOM element was discussed in #43, and it was decided that Turbolinks should stick to swapping out the `<body>` to maintain simplicity. Thanks again.

ianwhite:
I have the same use case as @metaskills.  

Because of the way browsers treat iFrames, it's not possible to move them out and back into the body after replace [see this SO answer](https://stackoverflow.com/questions/8318264/how-to-move-an-iframe-in-the-dom-without-losing-its-state#answer-8318401)

This means you need to setup/teardown any such scripts on each turbolinks body replacement, but there are often no APIs for this.

For example zendesk has API methods for SPAs (like changing the current URL and auth info) but has no setup/teardown api.

A `data-turbolinks-root` attribute, which is opt-in, would solve my issue, and possibly some of the issues above.  It would also not affect the default behaviour.

Is it possible to reopen this issue?

ahaurw01:
@ianwhite if it helps, we've been using the latest commit of https://github.com/spanishdict/turbolinks in production for the last several months which contains this feature and it has performed well for us. It is out of date with this repo by now and would require some work to get up to speed after the typescript rewrite.

ianwhite:
@ahaurw01 I've written my own fork, which is updated for TypeScript, but uses a configuration option to enable the feature.  https://github.com/ianwhite/turbolinks

wkasem:
@ianwhite how can i install your version of turboklinks through npm ?

ianwhite:
> @ianwhite how can i install your version of turboklinks through npm ?

Hi @wkasem the npm repository is @ian2w/turbolinks

wkasem:
> @ian2w

Sorry but can't find it :/
https://www.npmjs.com/search?q=%40ian2w%2Fturbolinks

ianwhite:
Apologies, here it is: https://www.npmjs.com/package/@ian2w/turbolinks

wkasem:
> Apologies, here it is: https://www.npmjs.com/package/@ian2w/turbolinks

Thanks a lot

ianwhite:
I've opened the issue tracker on https://github.com/ianwhite/turbolinks so we can stop spamming this thread.

PR #402, opened 2018-07-13
Update README.md

bdavidxyz:
See https://github.com/turbolinks/turbolinks-rails/pull/41


domchristie:
Thanks for this. I wonder if this might be better suited to the README in https://github.com/turbolinks/turbolinks-rails. What do you think?

bdavidxyz:
Thanks for the advice. However the whole docs are here so it make more easy and consitent for the developer to find it here. As you wish :)

PR #404, opened 2018-07-24
Modernize tests

sstephenson:
This branch addresses intermittent test failures in CI by replacing the existing [Blade](https://github.com/javan/blade)/[Replicant](https://github.com/sstephenson/replicant) `<iframe>`-based test suite with a modern stack using [Intern](https://theintern.io) and the [WebDriver interface](https://www.w3.org/TR/webdriver/). 

[Test cases](https://github.com/turbolinks/turbolinks/tree/tests/test/suites) are now written in TypeScript with `async`/`await`, flattening the nested hierarchy of callbacks.

The tests run in a Node.js process which controls browsers directly. (Currently the runner is only configured to test against a headless Chrome instance; we need to restore the Sauce Labs configuration before merging.)


sstephenson:
Porting the tests to other browsers is a bit more involved than I thought. We’ll revisit shortly.

PR #420, opened 2018-09-28
Fix Scrolling Issues

### LINKED ISSUE #181: Unable to avoid Scroll Position Lock on page transition
Rails App. I pared down all JS and left only Turbolinks.

What happens is when I scroll down y:350px, and click a link to another turbolinks enabled page, page2 loads at y:350px. 

When I turn it off. None of that happens. I didn't realize this was happening for probably half of the our app's life because our pages were all shorter than the fold. But once the content expanded past the fold, then we started noticing this issue.

So, I downloaded the repo, un-uglified the source code, just tossed a bunch of debuggers, stepped through an entire code execution, and couldn't specifically pin down when the position values changed in the code. 

So, one of the things that I tried was adding `<meta name="turbolinks-cache-control" content="no-cache">` to my layout, and lo, that seemed to resolve this issue... sort of. When the page renders, I scroll down a bit from the top within a few seconds of rendering, and it bumps me back up to the top.

Literally no JS is on the page except for what's imported from the Turbolinks Gem. This is happening in at least FF & Chrome. My basic understanding is that the pre-rendered page that we're advancing to is scrolling to the position, even though it looks like in the code that it's trying to ensure that that's not happening.

I don't know if I can replicate this in a bare bones app (as I have another turbolinks app where this doesn't occur). But I'm unsure as to what particular differences would be causing this in the slightest.

### LINKED ISSUE #241: Scrolling makes content blink in Safari
There's a `position:fixed` element on the page that is a link. If I was to click it while at the very top of the page, it'd load the linked page perfectly, as expected (no blinking or anything). However, if I am to scroll down and click the same thing, the content on the page goes blank before rendering.

I notice this happening on Safari 10.0.3, macOS 10.12.3. On Chrome and Firefox this doesn't seem to be an issue. See gif (sorry for potato quality):

turbolinks-blinks

domchristie:
This is an attempt at fixing #181 and #241.

When a page is invalidated, either by tracked asset changes or via the `turbolinks-visit-control` directive, it is reloaded but the user is returned to the previous scroll position :/ This can be jarring when moving between long pages with different layouts—someone will click a link and be dumped some way down the page when it loads. This is discussed in #181, and I have outlined a couple of solutions in https://github.com/turbolinks/turbolinks/issues/181#issuecomment-423448708. This pull request aims to implement the "improved solution" in a more structured way.

First, we set the `scrollRestoration` flag to `manual` to indicate to the browser that we'll be handling scroll positions when navigating via the history. This seemingly impacts the scroll position on reload, too. It's great for this bug as it means that if a page is invalidated and reloaded, we do nothing. However, we also don't want to break the refresh button :/ So to fix this we store the scroll position in session storage, and restore it on reload.

Configuring `scrollRestoration` has the nice side-effect of fixing scrolling flicker when navigation Back/Forward #241. The downside is that for a complete solution we need to persist scroll values across full page loads (on `beforeunload`), through session storage, which I'm not 100% convinced about.

App which demonstrates the unexpected/buggy behaviour: https://secret-bayou-73715.herokuapp.com

App which demonstrates this fix: http://sheltered-reef-10818.herokuapp.com

Fixes #181
Fixes #241

---

`scrollRestoration` is not supported in IE/Edge, so I would welcome some testers in those browsers in particular. ~Also, I should add some automated tests for this, but I'm having some difficulty in setting up the recent test framework. (I think Intern starts the server then immediately stops with no error.)~ Thanks to @jason0x43 for his help on this. My old version of Java was not compatible with Intern.


domchristie:
App which demonstrates the unexpected/buggy behaviour: https://secret-bayou-73715.herokuapp.com

App which demonstrates the fix: http://sheltered-reef-10818.herokuapp.com

kinduff:
Love this. Please merge :+1:

panoply:
Can someone merge this? Really need.

domchristie:
@panoply feel free to checkout and build https://github.com/domchristie/turbolinks/tree/scroll_on_load and report any feedback. Thanks!

domchristie:
> for a complete solution we need to persist scroll values across full page loads (on `beforeunload`), through session storage, which I'm not 100% convinced about.

This might prove to be useful in more situations as we could determine how a page was loaded (initial page load, manual reload, page invalidated reload, or turbolinks load), and then annotate load events accordingly.

panoply:
@domchristie No problems my end Safari / Chrome / Firefox

jasonhr13:
Ran into this issue today, would love this to get merged.

domchristie:
According to [the spec](https://html.spec.whatwg.org/multipage/history.html#history-1), the `scrollRestoration` mode is added to each session history entry. This pull request sets the mode to `manual` on start up, and is then used in all entries. I wonder if it might be cleaner to set this property according to the situation? (or if that would even work?)

gueorgui:
It would be really nice to get this merged!

zokioki:
Running into this issue as well, a merge would be awesome 🙏

AndersonSKM:
Running into this issue here. 🤞 🙏

megeek:
also running into this...please merge! 👍

danjenuil:
Running to this issue too! Please have it merged!

rsl:
this seems to still be alive, right?

rsl:
it doesn't look like this is going to be merged in any time soon. is there a way to use a fork of this to solve the problem? since this is just the javascript and is loaded by a rubygem, i honestly have no idea how to do this. looks like turbolinks itself has just been declared as superceded by turbo, in fact. would appreciate very much if anyone can help me get this fixed on a [rails 5] application using forks. crossing fingers and tia.

herunan:
Can you please finish checking. It would great to have it merged! Thank you.

PR #421, opened 2018-09-28
Prevent scroll position persistence when view invalidated

domchristie:
A simpler fix for #181 /#248 (I thought I tested this when working on #420, but obviously not well enough!)

As is discussed in #181, when a view is invalidated the scroll position is persisted when the page is refreshed. This results in an odd experience, particularly when navigating between long pages with different assets.

To fix this, the `window.location.reload` `forceReload` flag is set to `true` to scroll to the top on reload:

>  Besides caching behaviour the forcedReload flag also impacts how some browsers handle the scroll position: ordinary reload tries to restore the scroll position after reloading the page, while in forced mode (when parameter is set to true) the new DOM gets loaded with scrollTop == 0.
> https://developer.mozilla.org/en-US/docs/Web/API/Location/reload

I'm not sure what "some browsers" means, but it has the desired effect in ~Safari~, ~Chrome~, and Firefox on macOS.

### DIFF
--- a/src/turbolinks/browser_adapter.coffee
+++ b/src/turbolinks/browser_adapter.coffee
@@ -57,4 +57,4 @@ class Turbolinks.BrowserAdapter
     clearTimeout(@progressBarTimeout)
 
   reload: ->
-    window.location.reload()
+    window.location.reload(true)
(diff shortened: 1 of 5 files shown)


windy:
force reload maybe not scroll page to the top, tested on Google Chrome Version 69.0.3497.100 (Official Build) (64-bit).

I have another PR to fix this: 
https://github.com/turbolinks/turbolinks/pull/422

domchristie:
> force reload maybe not scroll page to the top, tested on Google Chrome Version 69.0.3497.100 (Official Build) (64-bit).

I have retested and can confirm that this is unfortunately the case. As the description mentions, I had tested this a number of times, and it appeared to work. Today, it didn't. Let's just say that browser caching behaviour is inconsistent!

windy:
@domchristie Yes, it's annoyed behavior about different browsers. `window.location.href = window.location.href` maybe a right solution for this. My safari Version 11.1.1 (13605.2.8) also work right.

PR #422, opened 2018-09-30
fix unexpected scroll behavior when visit page has meta head with reload

windy:
There is an unexpect scroll behavior:

I have two kinds of pages with different layout( using ruby on rails ):  agent pages and visit pages.

agent pages layout like this:

```
meta name="turbolinks-cache-control" content="no-cache"
= stylesheet_link_tag 'application', media: 'all', 'data-turbolinks-track': 'reload'
= javascript_include_tag 'application', 'data-turbolinks-track': 'reload'
```

visit pages layout like this:

```
meta name="turbolinks-cache-control" content="no-cache"
= stylesheet_link_tag 'pages', media: 'all', 'data-turbolinks-track': 'reload'
= javascript_include_tag 'pages', 'data-turbolinks-track': 'reload'
```

when user click link that will goto visit page from the bottom of agent page, they will find the new visit page scroll to bottom unexpectedly.

After some research, I found `window.location.reload()` will cause page scroll  the "right" place automatically. But it should not be the right behavior at the case above.

So we can replace `window.location.reload()` with `window.location.href = window.location.href`, fix the bug.

Here is a risk for this modify: 
https://stackoverflow.com/questions/2405117/difference-between-window-location-href-window-location-href-and-window-location

### DIFF
--- a/src/turbolinks/browser_adapter.coffee
+++ b/src/turbolinks/browser_adapter.coffee
@@ -57,4 +57,4 @@ class Turbolinks.BrowserAdapter
     clearTimeout(@progressBarTimeout)
 
   reload: ->
-    window.location.reload()
+    window.location.href = window.location.href


domchristie:
Unfortunately this seems to break the "Back" behaviour on Safari 11.1.1. Upon pressing Back, the page briefly loads the previous page before returning to the current one.

windy:
Let me see,  I just tested on chrome that's ok.

windy:
@domchristie I just rebuild a new rails project, found this:

The official  ruby gem turbolinks 5.2.0 has the same problem at safari 11.1.1 as you show. but 5.1.0 works correctly.

I think there is something wrong betweent 5.1.0 and 5.2.0. but not this modify.

domchristie:
> The official ruby gem turbolinks 5.2.0 has the same problem at safari 11.1.1 as you show. but 5.1.0 works correctly.
> 
> I think there is something wrong betweent 5.1.0 and 5.2.0.

OK, this will need investigating further.

Are you able to help test https://github.com/turbolinks/turbolinks/pull/420?

windy:
Let's me try 👍

PR #431, opened 2018-10-18
Ensure WebComponents triggers `connectedCallback`

seanpdoyle:
Closes turbolinks/turbolinks#337

Background
---

[Custom Elements][mdn] are a great fit for Turbolinks. Unfortunately,
more often than not it's still required to use the `webcomponents.js`
Custom Elements [polyfill] to support browsers other than Chrome and
Safari 11.

This, in itself, is not much of a problem. However, I noticed that on
browsers that need to fall back to the polyfill, my custom elements'
`connectedCallback` methods are not called after following a link,
breaking them all.

Problem
---

When visiting a new page, Turbolinks currently replaces the body by
setting `document.body`:

```coffee
assignNewBody: ->
  document.body = @newBody
```

However, the polyfill cannot detect document.body changes
(webcomponents/custom-elements#126). So it will not be able to walk the
DOM and call the `connectedCallback` methods of all custom elements.

Solution
---

Previously, `Turbolinks` replaced the document body by setting
`document.body`. However, the reference Custom Elements [polyfill]
does not patch the `document.body` property, the `connectedCallback` of
any custom elements in the new body will never be called.

By using `document.documentElement.replaceChild()`, we make it possible
for the Custom Elements polyfill to detect that new elements were
connected do the DOM.

[mdn]: https://developer.mozilla.org/en-US/docs/Web/Web_Components/Custom_Elements
[polyfill]: https://github.com/webcomponents/custom-elements

Co-authored-by: Sean Doyle <sean.p.doyle24@gmail.com>
Co-authored-by: Denis Washington <denis@denisw.de>


kmmbvnr:
Till this been fixed, I'm using modified custom-elements.js

I've added just after `PatchElement(internals)`

```js
  window.addEventListener('turbolinks:before-render', (event) => {
    internals.disconnectTree(document.body);
    internals.connectTree(event.data.newBody);
  });
```

sstephenson:
Thank you! https://github.com/turbolinks/turbolinks/releases/tag/v5.3.0-beta.1

PR #445, opened 2019-02-12
Port Turbolinks to TypeScript

sstephenson:
This pull request migrates the Turbolinks source code from CoffeeScript to TypeScript and ES modules. We plan to release this as version 5.3.0.

**Why?**
Basecamp's sister framework, [Stimulus](https://stimulusjs.org), is written in TypeScript, and we [recently ported](https://github.com/turbolinks/turbolinks/pull/404) the Turbolinks test suite to TypeScript. This branch unifies our front-end framework development stack ahead of work on Turbolinks 6.

**Do I need to change anything in my application?**
Most applications won't require any changes. If you have monkey-patched Turbolinks classes or methods, those patches may no longer work. However, the [public API](https://github.com/turbolinks/turbolinks#api-reference) is unchanged.

**How can I test the port?**
We have published a pre-release development version to npm as [5.3.0-beta.0](https://www.npmjs.com/package/turbolinks/v/5.3.0-beta.0), which you may reference in your `package.json` file. If you are using Sprockets, you can download the [dist file from unpkg](https://unpkg.com/turbolinks@5.3.0-beta.0/dist/turbolinks.js) and place it in your load path.

**What about existing pull requests?**
Existing pull requests will need to be rewritten, but this is largely a syntactic exercise. Most classes and methods are the same.

**Whither CoffeeScript?**
CoffeeScript has served us well [since 2012](https://github.com/turbolinks/turbolinks-classic/commit/879db1489156ed610e74f1befbb67ab5db6ee81e#diff-1ab071e0128cb6cfc0ad9e448d9aba7b) and will always hold a special place in this author's heart. TypeScript brings an improved editing experience (to easily jump to call sites or definitions, or rename classes and methods) and explicit interfaces (to specify the API between Turbolinks and its [native adapters](https://github.com/turbolinks/turbolinks/blob/aae02bc7e70d65598407636ec58592d6977d5147/src/adapter.ts), for example).


NurdTurd:
Hey @sstephenson,

I have a question. Does this result in any noticeable performance increases/decreases?

I understand TypeScript in the end is transpiled to JS so there should likely be not but was just wondering.

PR #446, opened 2019-02-15
Allow preventDefault() on "turbolinks:request-end" events

inxilpro:
This provides a way to cancel Turbolinks requests after the response is received, but before Turbolinks handles the response. A good example might include allowing JSON responses if you're on a page that supports client-side rendering. This is inspired, in part, by [a recent article outlining server-side apps that use client-side rendering](https://reinink.ca/articles/server-side-apps-with-client-side-rendering).

```javascript
document.addEventListener('turbolinks:request-start', function(event) {
  if ('renderApp' in window) {
    // Our client-rendered app is booted, so tell Turbolinks to accept JSON responses
    event.data.xhr.setRequestHeader('Accept', "text/html, application/xhtml+xml, application/json");
  }
});

document.addEventListener('turbolinks:request-end', function(event) {
  const xhr = event.data.xhr;
  if (xhr.getResponseHeader('Content-Type').includes('application/json')) {
    // We got JSON back, so stop Turbolinks from handling, and pass the data to our app instead
    event.preventDefault();
    const ok = xhr.status >= 200 && xhr.status < 300;
    window.renderApp(ok ? JSON.parse(xhr.responseText) : { error: 'Error loading response.', xhr });
  }
});
```

### DIFF
--- a/src/http_request.ts
+++ b/src/http_request.ts
@@ -100,7 +100,7 @@ export class HttpRequest {
   }
 
   notifyApplicationAfterRequestEnd() {
-    dispatch("turbolinks:request-end", { data: { url: this.url, xhr: this.xhr } })
+    return dispatch("turbolinks:request-end", { data: { url: this.url, xhr: this.xhr } })
   }
 
   // Private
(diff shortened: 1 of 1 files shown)


inxilpro:
It's also worth noting that Turbolinks will still push the location to history. We're really just preventing it from handling the response. It may make sense to pull the logic around the `Turbolinks-Location` header up so that it runs regardless of whether `request-end` is cancelled, but I held off on that because it's a bit more of an impactful change.

PR #456, opened 2019-03-16
Adds Turbolinks.setRootSelector feature

ianwhite:
Based on @ahaurw01 data-turbolinks-root work [#400] [#401]
Updated for the Typescript rewrite, with documentation.
Rewritten as a configuration option, so that it is opt-in, and slightly more performant than always querying for a "[data-turbolinks-root]" selector when it's not required.


ianwhite:
Oops, I didn't add some files - reopening a new pull request

Issue #184, opened 2016-09-29
Have you thought about using virtual DOM diffing?


ezekg:
(Preface: I know, I know… a virtual DOM? It might be a little overkill and probably not even close to being on the roadmap, but here me out anyways just for the sake of discussion.)

While developing a site a few weeks back, I wanted button hover states/animations to persist between page transitions. Using vanilla Turbolinks, I was experiencing 'jumpy' transitions (gif below). Although I could have used something like React to accomplish it, I thought that was a bit overkill for such a simple site. So, I figured I'd try out using Turbolinks with a virtual DOM.

I've been experimenting with a small monkey patch using the nifty [virtual-dom](https://github.com/Matt-Esch/virtual-dom) and [vdom-virtualize](https://github.com/marcelklehr/vdom-virtualize) libraries compiled with [browserify-rails](https://github.com/browserify-rails/browserify-rails):

```coffee
#= require turbolinks

diff = require "virtual-dom/diff"
patch = require "virtual-dom/patch"
virtualize = require "vdom-virtualize"

Turbolinks.SnapshotRenderer::assignNewBody = ->

  # Virtualize the current body
  curBody = virtualize document.body

  # Virtualize the new body
  newBody = virtualize @newBody

  # Diff the 2 virtual DOMs, creating a set of patches
  patches = diff curBody, newBody

  # Apply the resulting patches to the current DOM
  patch document.body, patches
```

It's obviously quite hacky, but it's enough to experiment with. And however hacky it may be, the results are pretty awesome. See the gifs below:

### Before
![](http://i.giphy.com/l0HlVHBNNBNZ7MMda.gif)

### After
![](http://i.giphy.com/l0HlwD0imF5ymaDjW.gif)

I thought I'd see what the thoughts were on something like this and maybe start a discussion on whether or not this kind of feature would be useful to other Turbolinks users.

sstephenson:
Hey @ezekg, thanks for bringing this up. I love how small your implementation is. 

I don’t see this coming to Turbolinks core anytime soon, but what I *would* really like to do is add hooks to the render events so you can implement a custom renderer yourself without monkey-patching. Probably something along the lines of calling `event.preventDefault()` in `turbolinks:before-render`, and then calling back into Turbolinks when your render is complete.

Please feel free to open a pull request if you’re interested in exploring custom rendering hooks.

nateberkopec:
[...]

PR #457, opened 2019-03-16
Adds Turbolinks.setRootSelector feature

ianwhite:
Based on @ahaurw01 data-turbolinks-root work [#400] [#401]
Updated for the Typescript rewrite, with documentation.
Rewritten as a configuration option, so that it is opt-in, and slightly more performant than always querying for a [data-turbolinks-root] selector.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ The npm package alone does not provide server-side support for Turbolinks redire
 - [Reloading When Assets Change](#reloading-when-assets-change)
 - [Ensuring Specific Pages Trigger a Full Reload](#ensuring-specific-pages-trigger-a-full-reload)
 - [Setting a Root Location](#setting-a-root-location)
+- [Changing the Root Replacement Element from `<body>`](#changing-the-root-replacement-element-from-body)
 - [Following Redirects](#following-redirects)
 - [Redirecting After a Form Submission](#redirecting-after-a-form-submission)
 - [Setting Custom HTTP Headers](#setting-custom-http-headers)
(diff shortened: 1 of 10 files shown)


ianwhite:
I know that a similar feature has been proposed and rejected in #43 and #401. I've tried to write the new feature so that it takes more work to opt-in to it (you need to configure the namespace).

This is my first experience of TypeScript. I've tried the follow the style of the codebase, but will take any pointers gladly.

The tests that I wrote slightly departed from the style in that I used the 'turbolinks:load' event to check when the page had changed, rather than the `nextBody` getter.  This is because the implementation of the `nextBody` getter relies on the premise that the body element is swapped out.  If this pull request is accepted, work might need to be done to make that clearer.

Many thanks for Turbolinks!

ianwhite:
This can be tried out by installing from npm @ian2w/turbolinks

jnstq:
This would be great for my use case with the spotify playback api, that inserts a iframe.  Permanent do not works as the iframe gets reloaded on the body replacement.

jnstq:
I tried this and it exactly what I need to keep the iframes between pages, but it breaks preview and navigate back. Would be great to have this feature, but at the moment this https://github.com/turbolinks/turbolinks/issues/184#issuecomment-451688212 works fine with turbolinks 5.2

markets:
Hi 👋 this would be amazing! 👌 

Turbolinks v5 breaks (even with `data-turbolinks-permanent`) plugins like [Olark](https://www.olark.com) and this change can help a lot to make it work easily (by just replacing a 1-level deeper DIV for example, as we are doing with `pjax`).

AlexLarra:
It would be great to have this feature for easily adapt 3rd party js. Is there any possibility to be merged?

PR #495, opened 2019-10-02
Fix default browser refresh behaviour after POST request

### LINKED ISSUE #229: Rails: unexpected behaviour in Chrome and Firefox when refreshing invalid form page
Original posted on [SO:](http://stackoverflow.com/questions/39742354/rails-unexpected-behaviour-in-chrome-and-firefox-when-refreshing-invalid-form-p/41744072#41744072) 

"When a user refreshes the page after submitting an invalid form, the browser should attempt to resubmit the form. In Chrome and Firefox (but not Safari) the browser performs a GET request on the create/update url instead.

The browser should resubmit the form (usually asking "Page reare you sure?" or similar), and render the same form-with-errors page. In Chrome and Firefox (but not Safari), on page refresh the browser sends a GET request to the POST url, effectively calling index.

f the url is not RESTful, or if you just happen to have no index or show views, reloading causes the app to throw a "No route" error."

 This will lead a bad experience for user who try to fix their input on the form by pressing refresh button.

Such behavior only happen on Turbolink 5 , when the application using TL 2.5.3 it behave as expected.

domchristie:
Turbolinks initialisation currently breaks the default refresh behaviour when it's installed after a POST submission. A POST followed by a refresh should trigger an "Are you sure…?" confirm dialog, whereas with Turbolinks installed, it just reloads the existing location with a GET.

When starting up, [Turbolinks stores the initial location and restoration identifier in the session history](https://github.com/turbolinks/turbolinks/blob/b85b343f554371bea9aee4d2165b99a2b0522211/src/controller.ts#L101) by calling `replaceState`. This standardises the way to retrieve a restoration identifier on any popstate event. However, according the to the spec:

> 3. [If the state push flag is not set], update the current entry in browsingContext's session history so that … it represents a GET request, if it currently represents a non-GET request (e.g. it was the result of a POST submission).
> https://html.spec.whatwg.org/multipage/history.html#history-1

So when a page loads after a POST, Turbolinks replaces the current entry with a GET to the same location, and therefore will not re-POST or display a confirm dialog upon refresh\*.

This pull request tweaks initialisation to store the initial restoration identifier and location as properties on the `History` instance, avoiding the need to use `replaceState`. When the history is popped to the initial entry, `event.state` will be `null`, and by asserting that the existing location matches the stored initial location, the restoration identifier is retrieved accordingly. Pops to non-initial entries will be handled as before.

Fixes #229

---

\* This is observed in Chrome and Firefox, which comply with the spec. Safari behaves as we _want_, but does not comply with the spec.

### DIFF
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -97,8 +97,7 @@ export class Controller {
   startHistory() {
     this.location = Location.currentLocation
     this.restorationIdentifier = uuid()
-    this.history.start()
-    this.history.replace(this.location, this.restorationIdentifier)
+    this.history.start(this.location, this.restorationIdentifier)
   }
 
   stopHistory() {
--- a/src/history.ts
+++ b/src/history.ts
@@ -9,6 +9,8 @@ type HistoryMethod = (state: any, title: string, url?: string | null | undefined
 
 export class History {
   readonly delegate: HistoryDelegate
+  initialLocation!: Location
+  initialRestorationIdentifier!: string
   started = false
   pageLoaded = false
 
@@ -16,8 +18,11 @@ export class History {
     this.delegate = delegate
   }
 
-  start() {
+  start(location: Location, restorationIdentifier: string) {
     if (!this.started) {
+      this.initialLocation = location
+      this.initialRestorationIdentifier = restorationIdentifier
+
       addEventListener("popstate", this.onPopState, false)
       addEventListener("load", this.onPageLoad, false)
       this.started = true
@@ -44,10 +49,10 @@ export class History {
 
   onPopState = (event: PopStateEvent) => {
     if (this.shouldHandlePopState()) {
-      const { turbolinks } = event.state
-      if (turbolinks) {
-        const location = Location.currentLocation
-        const { restorationIdentifier } = turbolinks
+      const location = Location.currentLocation
+      const restorationIdentifier = this.restorationIdentifierForPopState(event)
+
+      if (restorationIdentifier) {
         this.delegate.historyPoppedToLocationWithRestorationIdentifier(location, restorationIdentifier)
       }
     }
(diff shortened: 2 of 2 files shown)


jclusso:
Is this going to get merged?

akiraandy:
This would be great to have merged in... This would fix an issue I've been struggling with.

domchristie:
@jclusso @akiraandy this pull request has been updated with the latest changes from the master branch, so if you are interested in trying it out, you may wish to download and build from the source

jclusso:
@domchristie is this going to get merged anytime soon or are you waiting on one of us to try it out?

domchristie:
@jclusso I don't think there are plans to merge this soon. From [the sounds of things](https://twitter.com/dhh/status/1257717471995236352), there might be some [tasty updates](https://twitter.com/search?q=from%3Adhh%20herbs%20spices&src=typed_query&f=live) coming once Basecamp launch hey.com.

dsazup:
Just run into this. Would be nice to get it merged. Is there an easy way to monkey patch this without running a fork? 🤔

judy:
Pulled this down and tried it in our project. It fixes the exact problem we were having! 😄 But without running a fork, we might have to figure out how to monkey patch this ourselves in the meanwhile...

martinbarilik:
Is this why using bootstrap-table and turbolinks 5 the F5 button won't fire before-cache trigger and table gets corrupted, because it's initialized again without being destoyed? 

And when pressing button with POST request which redirects to the same view wont fire the trigger as well and corrupts the table ?

domchristie:
Hey @martinbarilik refreshing a page and submitting standard (non-AJAX) forms won't trigger `turbolinks:before-cache` because they are not handled by Turbolinks. In these cases the page fully reloads so any previously initialised JavaScript libraries will be naturally destroyed. You may want to check that bootstrap-table is loaded and initialised properly on standard page loads.

martinbarilik:
@domchristie thanks for reply. How am i supposed to check it on standard page load? I need to have a full controll over the library being destroyed and loaded again. Otherwise it's mess ...

PR #523, opened 2020-02-25
Handle onPopState when turbolinks is disabled

bnjamin:
When clicking on a link where Turbolinks is disabled via the `data-turbolinks="false"` annotation the PopStateEvent would have no state. This would result in an `Cannot read property 'turbolinks' of null` error.
It is now handled by doing an early return in that situation.

If we were using TS version 3.7 we could've used optional chaining instead of early return guards.

### DIFF
--- a/src/history.ts
+++ b/src/history.ts
@@ -43,13 +43,14 @@ export class History {
   // Event handlers
 
   onPopState = (event: PopStateEvent) => {
-    if (this.shouldHandlePopState()) {
-      const { turbolinks } = event.state
-      if (turbolinks) {
-        const location = Location.currentLocation
-        const { restorationIdentifier } = turbolinks
-        this.delegate.historyPoppedToLocationWithRestorationIdentifier(location, restorationIdentifier)
-      }
+    if (!this.shouldHandlePopState()) return
+    if (!event.state) return
+
+    const { turbolinks } = event.state
+    if (turbolinks) {
+      const location = Location.currentLocation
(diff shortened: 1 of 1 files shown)


bnjamin:
This was handled in the coffeescript version

javan:
> This was handled in the coffeescript version

Confirmed! :) https://github.com/turbolinks/turbolinks/blob/ef969ab53e6bb1d2c0be37f6d8dec38af8e498b2/src/turbolinks/history.coffee#L27-L28

src/history.ts
```diff
@@ -43,13 +43,14 @@ export class History {
...
-      }
+    if (!this.shouldHandlePopState()) return
+    if (!event.state) return
+
+    const { turbolinks } = event.state
+    if (turbolinks) {
+      const location = Location.currentLocation
+      const { restorationIdentifier } = turbolinks
+      this.delegate.historyPoppedToLocationWithRestorationIdentifier(location, restorationIdentifier)
     }
```
javan: Let's keep the existing no-early-return style or use it consistently:
```suggestion
    const { turbolinks } = event.state
    if (!turbolinks) return
    
    const location = Location.currentLocation
    const { restorationIdentifier } = turbolinks
    this.delegate.historyPoppedToLocationWithRestorationIdentifier(location, restorationIdentifier)
```

bnjamin: 👍

javan:
Thanks!

PR #530, opened 2020-03-23
New feature: data-turbolinks-permanent="shallow"

gilbert:
This allows an element to maintain object identity but still have its contents updated by the server. Useful for animating layout elements.


gilbert:
Huh, I just removed `data-turbolinks-permanent="shallow"` from my code entirely and it still worked. I must have inadvertently fixed my issue some other way.

PR #541, opened 2020-05-04
Mention that jQuery's `on` will not see event.data

MaxLap:
jQuery wraps events into a home-made event object. This wrapping makes the event.data unavailable. This commit adds a little mention of that fact to help a future developer save time and frustration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -499,7 +499,8 @@ Detects whether Turbolinks is supported in the current browser (see [Supported B
 
 ## Full List of Events
 
-Turbolinks emits events that allow you to track the navigation lifecycle and respond to page loading. Except where noted, Turbolinks fires events on the `document` object.
+Turbolinks emits events that allow you to track the navigation lifecycle and respond to page loading. Except where noted, Turbolinks fires events on the `document` object.  
+Note that `event.data` (when offered) is not available if you use jQuery's `$(document).on` method. Prefer `document.addEventListener`.
 
 * `turbolinks:click` fires when you click a Turbolinks-enabled link. The clicked element is the event target. Access the requested location with `event.data.url`. Cancel this event to let the click fall through to the browser as normal navigation.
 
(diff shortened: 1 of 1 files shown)


MaxLap:
I debated that, but I thought that was maybe going too deep into jQuery's details for a package that doesn't rely on jQuery. I can change it

domchristie:
[`$event.originalEvent` is already mentioned in the README](https://github.com/turbolinks/turbolinks#canceling-visits-before-they-start), so there is a precedent. It might be useful to also mention it in the events section.

MaxLap:
I personally don't get why this says `$event.originalEvent` instead of `event.originalEvent`. I would be tempted not to have the `$`.

Should I mention it once at the top like I did, or go through each event and mention `event.originalEvent.data.newbody` ?

domchristie:
> I personally don't get why this says $event.originalEvent instead of event.originalEvent. I would be tempted not to have the $.

`$` reinforces that it's a jQuery-ified event, and not a native one.

> Should I mention it once at the top like I did, or go through each event

I think just once should be fine 👍

MaxLap:
Updated the note as suggested.

README.md
```diff
@@ -499,7 +499,8 @@ Detects whether Turbolinks is supported in the current browser (see [Supported B
 
 ## Full List of Events
 
-Turbolinks emits events that allow you to track the navigation lifecycle and respond to page loading. Except where noted, Turbolinks fires events on the `document` object.
+Turbolinks emits events that allow you to track the navigation lifecycle and respond to page loading. Except where noted, Turbolinks fires events on the `document` object.  
+Note that when using jQuery, the data on the event must be accessed as `$event.originalEvent.data.url`.
```
domchristie, line 503: One small fix:

```suggestion
Note that when using jQuery, the data on the event must be accessed as `$event.originalEvent.data`.
```

MaxLap: Good catch, thank you!

domchristie:
Thanks @MaxLap!

